Issue #378, opened 2015-01-06
Use a data validation framework


jaredkerim:
The current validation code is written from scratch rather than using some kind of validation framework.  This makes it difficult to reason about how a given field is validated or track down where a validation failure is occurring.  There are many well established frameworks for validating data, we could use the SQLAlchemy model level validation, something pyramid provides, or something more standalone like:

https://github.com/schematics/schematics

http://wtforms.readthedocs.org/en/latest/

hannosch:
Or we can use the one we already use in more places -> http://docs.pylonsproject.org/projects/colander/en/latest/

We do need a validation framework that is independent of any storage models, as we validate data as part of the data pipeline when all we have is blobs of JSON to work with.

jaredkerim:
Colander it is!

PR #387, opened 2015-01-16
Use colander for data validation #378

jaredkerim:
- Add schemas for Measure, Cell, CellMeasure, Wifi, WifiMeasure
- Split tests into more isolated tests for Wifi
- Leave normalize_ layer intact, can update tasks to call into Schemas directly later
- Update tests as necessary


ichnaea/scripts/importer.py
```diff
@@ -44,8 +43,9 @@ def load_file(session, source_file, batch_size=1000, userid=None):
                     # convert from unixtime to utc
                     time = datetime.datetime.utcfromtimestamp(time)
 
-                key = normalized_wifi_key(str(fields[1]))
-                if not valid_wifi_pattern(key):  # pragma: no cover
+                try:
+                    key = WifiKeyNode(String()).deserialize(str(fields[1]))
```
jaredkerim: I'm not a big fan of having to explicitly instantiate this validator like this (needing to import String from colander).  We should be able to use it by itself without needing to import additional things from colander.  I'm not sure what the best way around this is.

jaredkerim: I do think, however, that eventually we should call the colander validators directly instead of using the normalize_thing_dict interfaces.  I left them intact for now because we can worry about refactoring those bits later.

hannosch: I think we probably want to define an actual complete schema for the import data csv format here. If we have a full schema, this code might get less ugly - but that's a future todo.

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+    """
+    A node containing a valid report_id.
+    ex: 489cc8dc9d3d11e4a87d02442b52e5a0
+    """
+
+    def preparer(self, cstruct):
+        return cstruct or uuid.uuid1().hex
+
+
+class TimeNode(SchemaNode):
```
hannosch: Maybe call it a DateNode? It doesn't contain a time part.

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+
+    def preparer(self, cstruct):
+        return normalized_wifi_key(cstruct)
+
+    def validator(self, node, cstruct):
+        if not valid_wifi_pattern(cstruct):
+            raise Invalid(node, 'Invalid wifi key')
+
+
+class ReportNode(SchemaNode):
```
hannosch: I'd prefer ReportIdNode, a "report" is a full schema with lots of nodes in it, not just a single node.

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+    """
+    lat = SchemaNode(Float(), missing=0.0, validator=Range(MIN_LAT, MAX_LAT))
+    lon = SchemaNode(Float(), missing=0.0, validator=Range(-180, 180))
+    accuracy = DefaultNode(
+        Float(), missing=0, validator=Range(0, MAX_ACCURACY))
+    altitude = DefaultNode(
+        Float(), missing=0, validator=Range(MIN_ALTITUDE, MAX_ALTITUDE))
+    altitude_accuracy = DefaultNode(
+        Float(), missing=0, validator=Range(0, MAX_ALTITUDE_ACCURACY))
+    heading = DefaultNode(Float(), missing=-1, validator=Range(0, MAX_HEADING))
```
hannosch: accuracy, altitude and altitude_accuracy should all be integers. The db fields are ints and the missing and ranges are all specified as ints.

ichnaea/data/schema.py
hannosch: Oh and actually maybe call it something more specific like a RecentMonthNode - the normalized_time function changes any incoming time to the first of the month of the specified date and magically transforms and future or too-far past dates (more than 2 months in the past) into "now".

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+
+class ValidWifiSchema(ValidMeasureSchema):
+    """
+    A Schema which validates the fields present in a a wifi measurement.
+    """
+    signal = DefaultNode(Integer(), missing=0, validator=Range(-200, -1))
+    signalToNoiseRatio = DefaultNode(
+        Integer(), missing=0, validator=Range(0, 100))
+    key = WifiKeyNode(String())
+    channel = SchemaNode(Integer(), validator=Range(0, 166))
```
hannosch: maybe call out missing=0 explicitly here for the channel node?

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+            raise Invalid(
+                schema, 'Check against the list of all known valid mccs')
+
+        if (data['radio'] == RADIO_TYPE['cdma']
+                and (lac_missing or cid_missing)):
+            raise Invalid(schema, (
+                'Skip CDMA towers missing lac or cid '
+                '(no psc on CDMA exists to backfill using inference)'))
+
+        radio_types = RADIO_TYPE['gsm'], RADIO_TYPE['umts'], RADIO_TYPE['lte']
```
hannosch: could you put brackets around this to make it an obvious tuple?

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+    total_measures = SchemaNode(Integer(), missing=0)
+    range = SchemaNode(Integer(), missing=0)
+
+
+class ValidCellMeasureSchema(ValidCellBaseSchema):
+    """
+    A Schema which validates the fields present in
+    all Cell measurements.
+    """
+    created = SchemaNode(String(), missing=None)
```
hannosch: created as a string node looks odd, why isn't this a datetime node?

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+    def deserialize(self, data):
+        if data:
+            # Sometimes the asu and signal fields are swapped
+            if data.get('asu', 0) < -1 and data.get('signal', None) == 0:
+                data['signal'] = data['asu']
+                data['asu'] = self.fields['asu'].missing
+
+        return super(ValidCellMeasureSchema, self).deserialize(data)
+
+    def validator(self, schema, data):
```
hannosch: This looks a lot like the ValidCellBaseSchema.validator. That one has the extra check for mcc against the whitelist, which wouldn't hurt to do here either.

So maybe call `super(ValidCellMeasureSchema, self).validator(schema, data)` first and then do the in_country check afterwards?

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+# Utility functions
+
+def normalized_time(time):
+    """
+    Takes a string representation of a time value, validates and parses
+    it and returns a JSON-friendly string representation of the normalized
+    time.
+    """
```
hannosch: This docstring was bad before, but could you add a sentence to mention the whole "convert to first of month" and "magically convert dates older than two months to now" business?

ichnaea/data/tests/test_validation.py
```diff
@@ -72,34 +63,18 @@ def make_cell_submission(self, **kw):
...
-                       ('wimax', -1),
-                       ('', -1),
-                       ('hspa', -1),
-                       ('n/a', -1)]
+        self.radio_pairs = [
+            ('gsm', 0),
+            ('cdma', 1),
+            ('umts', 2),
+            ('wcdma', 2),
+            ('lte', 3),
```
hannosch: Maybe use the `RADIO_TYPE['gsm']` for the valid pairs here?

ichnaea/data/tests/test_validation.py
```diff
@@ -256,7 +231,161 @@ def test_asu_signal_field_mix_up(self):
         (measure, cell) = self.make_cell_submission(asu=-75, signal=0)
         self.check_normalized_cell(measure, cell, {'signal': -75})
 
-    def test_normalize_wifis(self):
+    def test_cid_65535_without_a_valid_lac_sets_cid_to_invalid(self):
+        (measure, cell) = self.make_cell_submission(cid=65535, lac=-1, psc=1)
+        self.check_normalized_cell(measure, cell, {'cid': -1})
+
+    def test_CDMA_cell_records_must_have_MNC_MCC_LAC_CID(self):
```
hannosch: uppercase letters look weird to me here, I think we lowercased cdma, mcc, etc in test names. Also maybe use `test_cdma_records_must_have_complete_id`, or at least sort mcc, mnc in their usual order :)

ichnaea/data/tests/test_validation.py
```diff
@@ -256,7 +231,161 @@ def test_asu_signal_field_mix_up(self):
...
+                radio='cdma', **entry[0])
+            self.check_normalized_cell(measure, cell, entry[1])
+
+    def test_records_fail_the_mcc_check(self):
+        entries = [
+            {'mcc': 0, 'mnc': 2, 'lac': 3, 'cid': 4},
+            {'mcc': -1, 'mnc': 2, 'lac': 3, 'cid': 4},
+            {'mcc': -2, 'mnc': 2, 'lac': 3, 'cid': 4},
+            {'mcc': 2000, 'mnc': 2, 'lac': 3, 'cid': 4},
+
```
hannosch: extra blank line

ichnaea/data/tests/test_validation.py
```diff
@@ -256,7 +231,161 @@ def test_asu_signal_field_mix_up(self):
...
+        ]
+
+        for entry in entries:
+            if 'radio' not in entry:
```
hannosch: we can remove that if here

ichnaea/data/schema.py
```diff
@@ -0,0 +1,376 @@
...
+    Boolean,
+    DateTime,
+    Float,
+    Integer,
+    Invalid,
+    MappingSchema,
+    Range,
+    SchemaNode,
+    String,
+    iso8601
```
hannosch: add a trailing comma

ichnaea/data/tests/test_validation.py
```diff
@@ -45,18 +48,6 @@ def check_normalized_cell(self, measure, cell, expect):
...
-
-        if expect is None:
-            self.assertEqual(result, expect)
-        else:
-            for (k, v) in expect.items():
-                self.assertEqual(result[k], v)
-        return result
-
     def make_cell_submission(self, **kw):
         measure = dict(radio='umts',
```
hannosch: Actually a top-level radio entry of umts wouldn't be valid. Top level is meant to be the phone type, which can only be '', 'gsm' or 'cdma'.

ichnaea/data/tests/test_validation.py
```diff
@@ -256,7 +231,161 @@ def test_asu_signal_field_mix_up(self):
...
+
+    def test_fails_because_it_has_an_MNC_above_1000_for_a_GSM_network(self):
+        entries = [
+            {'mcc': FRANCE_MCC, 'mnc': 1001, 'lac': 3, 'cid': 4,
+             'radio': 'gsm'},
+        ]
+
+        for entry in entries:
+            if 'radio' not in entry:
+                entry['radio'] = 'cdma'
```
hannosch: here we can just take away these two lines, as we specify radio explicitly in the above entry

ichnaea/data/tests/test_validation.py
```diff
@@ -256,7 +231,161 @@ def test_asu_signal_field_mix_up(self):
...
+            (measure, cell) = self.make_cell_submission(**entry)
+            self.check_normalized_cell(measure, cell, None)
+
+    def test_wrong_radio_type_is_corrected_for_large_cid(self):
+        measure, cell = self.make_cell_submission(cid=65536, radio='gsm')
+        self.check_normalized_cell(measure, cell, {'radio': 2})
```
hannosch: maybe use `RADIO_TYPE['umts']` here instead of the 2

ichnaea/data/validation.py
```diff
@@ -1,283 +1,52 @@
...
-            return None
-        n[k] = v
-
-    # copy forward anything not specified
-    for (k, v) in d.items():
-        if k not in n:
-            n[k] = v
-    return n
+        validated = ValidMeasureSchema().deserialize(data)
+    except Exception:
```
hannosch: Why catch a blanket Exception here? Later functions just catch Invalid it seems.

ichnaea/data/tasks.py
```diff
@@ -21,7 +21,7 @@
 )
 from ichnaea.data.validation import (
     normalized_measure_dict,
-    normalized_wifi_measure_dict,
+    normalized_wifi_dict,
     normalized_cell_measure_dict,
```
hannosch: This change looks odd, but seems to be a result of there not being a ValidWifiMeasureSchema. For cell stuff there is however a distinct ValidCellMeasureSchema. I'd rather have a distinct ValidWifiMeasureSchema class, even if it's just a placeholder, so the cell and wifi schemas and functions are consistent with each other.

hannosch:
Could you add any changes as additional commits to this PR? With so many tiny little changes it becomes easier to review additional commits instead of having to go through the entire rebased diff again. Once this is all ready and addressed, we can squash back all commits into one.

ichnaea/data/schema.py
jaredkerim: I have moved this to a separate ticket in #389

ichnaea/scripts/importer.py
jaredkerim: This has been moved to #390

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: This mirrors the existing behaviour where created is a string and isn't validated or checked against.  If we want to change this behaviour it should be addressed in a separate ticket.

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/schema.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/tests/test_validation.py
jaredkerim: +1

ichnaea/data/validation.py
jaredkerim: +1

ichnaea/data/tasks.py
jaredkerim: Moved into #391

hannosch:
Squashed the three commits into one (cb6b6af67e6cecb7b298dfa000d617396f8f0df6), merged it and addressed some more nitpicks in a separate commit.

Issue #291, opened 2014-08-07
Look at extended RNC values leading to a larger cid range for UMTS


hannosch:
Usually for UMTS we capture a combined RNC + Node-B cell id as the cid value and assume a maximum value of 268435455 for it (28 bits, 12 bit RNC + 16 bit cell id).

It appears there are some cases where the RNC part is actually larger and can be up to 16 bit in itself. Thus the combined value can be up to 32bit (4294967295).

We should accept that maximum value and make sure the DB field we use can store such values. We should also extend the range checks to be dependent on the radio type, so we don't allow these high values for GSM/CDMA networks.

Possible reference info at http://www.broadband-forum.org/cwmp/tr-196-1-0-0.html

hannosch:
This was done via #393.

PR #393, opened 2015-01-20
Update LAC and CID Validation #291

jaredkerim:
- Change the cid field in the cell, cell_measure and cell_blacklist table from a signed int to an unsigned int (4 byte in mysql, so exactly fits the 32bit values)
- Change the lac field in the same tables from a signed int to an unsinged small int (2 byte, so fits all the 16bit values)
- Change the "unknown" token for cid from -1 to 0
- Change the "unknown" token for lac from -1 to 0
- Change the valid range for UMTS cid to be 32bit, 1 to 2 ** 32 - 1
- Introduce new check for CDMA cid to be 16bit, 1 to 2 ** 16 - 1
- Introduce new check for LTE cid to be 28bit, 1 to 2 ** 28 - 1


E3V3A:
Good to know. Thanks.

Issue #370, opened 2014-12-17
Move new distance functions to geocalc.py


jaredkerim:
Move the functions in util.py:

https://github.com/mozilla/ichnaea/blob/master/ichnaea/util.py#L12-L28

into geocalc.py and document accordingly

hannosch:
Closed via #396.

PR #396, opened 2015-01-21
Move new distance functions to geocalc.py #370

jaredkerim:
- move functions from ichnaea.util to ichanea.geocalc
- move tests

r? @hannosch


hannosch:
r+, feel free to merge yourself.

If you have time, could you try to add this into the sphinx docs? We don't use them much, but I'd like to get better there over time. And things that already have doc strings are an easy win.

We have docs/internal_api/geocalc.rst which needs some new autofunction entries. Building docs locally should work via `make docs`, generated and viewable locally by opening `docs/build/html/internal_api/index.html`. In the end this all ends up on http://mozilla-ichnaea.readthedocs.org/en/latest/internal_api/index.html

jaredkerim:
docs updated!

Issue #400, opened 2015-01-28
Refactor locate service code


jaredkerim:
The code located at:

https://github.com/mozilla/ichnaea/blob/master/ichnaea/service/locate.py

Is designed partially to allow an extensible set of search types, but not in a very clean and modular way.  We should refactor this code to break the search logic into two parts: a SearchBroker, and a SearchProvider.  

A SearchProvider is responsible for performing one specific type of search, so for our current implementation that would be each of: cell, OCID cell, cell lac, wifi, geoip.  In the future we will add OCID cell lac (#351), as well as external API calls.  To prepare for that, refactoring the code now will make extending the number and types of search providers easier.

The SearchBroker will collect the search results of each SearchProvider, and either aggregate or decide which to use.  Presently the search logic uses a 'drill down' approach from the broadest to the narrowest search type and uses the best match it gets, if any.  When the SearchBroker and SearchProvider are clearly distinct form one another, changing how this logic behaves will be simpler without having to modify how each SearchProvider implements its specific type of search.

hannosch:
This has been merged and is all done. I've expanded a bit on it, unifying the geoip and database session based providers and introducing a stateful location result object. The result object made it possible to simplify the main loop by doing things like "new_result.more_accurate(old_result)", moving the logic on how to compare results onto these result objects.

In order to fix the last TODO (cell/cell_lac metrics) I had to add back some rather ugly code. This is going to get even more messy when we add the ocid_area provider. Maybe some other pattern is going to be obvious once we do that.

The overall design has already proven itself, as I managed to almost trivially add a new CellCountryProvider for providing cell key (mcc) based country estimates.

hannosch:
Actually there's a bit more general refactoring that could be done, that isn't specific to #351.

To make the logging work, I had to introduce a log_group name on each provider and do some grouped stats gathering at the end of the main locate loop. Maybe this could be generalized into the concept of a provider group, with one group per incoming data field. So we'd loop over three groups (geoip, cell, wifi) and one to multiple providers inside each group. That might make it more obvious to distinguish between a "cell_miss", "cell_hit" and "cell_lac_hit", as we can only determine that after we queried all cell based providers.

[...]

PR #403, opened 2015-02-13
Refactor search logic #400

jaredkerim:
- Make a clear distinction between an individual search provider (cell, cell lac, wifi, geoip) and a location searcher which combines them
- Add a test to hit the OCIDCell table
- Remove unnecessary stats logging

@hannosch r?  The Big One.

### DIFF
--- a/ichnaea/service/geolocate/tests.py
+++ b/ichnaea/service/geolocate/tests.py
@@ -57,7 +57,8 @@ class TestGeolocate(AppTestCase):
 
         self.check_stats(
             counter=[self.metric_url + '.200',
-                     self.metric + '.api_key.test']
+                     self.metric + '.api_key.test',
+                     self.metric + '.api_log.test.cell_hit']
         )
 
         self.assertEqual(res.content_type, 'application/json')
--- a/ichnaea/service/geolocate/tests.py
+++ b/ichnaea/service/geolocate/tests.py
@@ -57,8 +57,7 @@ class TestGeolocate(AppTestCase):
 
         self.check_stats(
             counter=[self.metric_url + '.200',
-                     self.metric + '.api_key.test',
-                     self.metric + '.api_log.test.cell_hit']
+                     self.metric + '.api_key.test']
         )
 
         self.assertEqual(res.content_type, 'application/json')
@@ -87,8 +86,7 @@ class TestGeolocate(AppTestCase):
                 ]},
             status=200)
         self.check_stats(
-            counter=[self.metric + '.api_key.test',
-                     self.metric + '.api_log.test.wifi_hit'])
+            counter=[self.metric + '.api_key.test'])
         self.assertEqual(res.content_type, 'application/json')
         self.assertEqual(res.json, {"location": {"lat": 1.001,
                                                  "lng": 1.002},
@@ -119,7 +117,6 @@ class TestGeolocate(AppTestCase):
         # Make sure to get two counters, a timer, and no traceback
         self.check_stats(
             counter=[self.metric + '.api_key.test',
-                     self.metric + '.api_log.test.wifi_miss',
                      self.metric_url + '.404'],
             timer=[self.metric_url],
         )
--- a/ichnaea/service/locate.py
+++ b/ichnaea/service/locate.py
@@ -2,7 +2,6 @@ from collections import defaultdict, namedtuple
 import operator
 
 import iso3166
-import mobile_codes
 from sqlalchemy.sql import and_, or_
 from sqlalchemy.orm import load_only
 
@@ -18,11 +17,10 @@ from ichnaea.data.validation import (
 )
 from ichnaea.geocalc import (
     distance,
-    location_is_in_country,
 )
 from ichnaea.logging import (
-    get_heka_client,
     get_stats_client,
+    get_heka_client,
     RAVEN_ERROR,
 )
 from ichnaea.models import (
@@ -65,26 +63,6 @@ def estimate_accuracy(lat, lon, points, minimum):
     return max(accuracy, minimum)
 
 
-def most_common_elements(iterable):
-    """
-    Given an iterable, return a list of elements from that iterable:
-        * If the iterable is empty, return an empty list.
-        * If there is one element that's most common in the iterable,
-          return a list with that element.
-        * If there are multiple elements which occur equally often,
-          return a list with all of those elements.
-    """
-    counts = defaultdict(int)
-    for e in iterable:
-        counts[e] += 1
-
-    if not counts:
-        return []
-
-    max_count = max(counts.values())
-    return [e for e, n in counts.items() if n == max_count]
-
-
 def map_data(data):
     """
     Transform a geolocate API dictionary to an equivalent search API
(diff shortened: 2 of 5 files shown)


ichnaea/service/geolocate/tests.py
```diff
@@ -57,8 +57,7 @@ def test_ok_cell(self):
 
         self.check_stats(
             counter=[self.metric_url + '.200',
-                     self.metric + '.api_key.test',
-                     self.metric + '.api_log.test.cell_hit']
```
hannosch: The `m.api_log.test.*_hit/miss` metrics are actually the useful ones, those should stay.

ichnaea/service/locate.py
```diff
@@ -18,11 +17,10 @@
 )
 from ichnaea.geocalc import (
     distance,
-    location_is_in_country,
 )
 from ichnaea.logging import (
-    get_heka_client,
     get_stats_client,
+    get_heka_client,
```
hannosch: In a squashed diff this looks silly :) Maybe reorder them alphabetically again.

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+        self.session = session
+        self.heka_client = get_heka_client()
+        super(AbstractLocationProvider, self).__init__(*args, **kwargs)
+
+    def locate(self, data):
+        raise NotImplementedError()
+
+
+class AbstractCellLocationProvider(AbstractLocationProvider):
+
```
hannosch: Maybe add a fake `__init__` or class attributes for the `self.` values used in this class, so it's obvious what a concrete implementation has to do? For example:
```
class AbstractCellLocationProvider(AbstractLocationProvider):

heka_client = None
models = ()
session = None
```

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+                    queried_objects.append(network)
+
+        return queried_objects
+
+    def locate(self, data):
+        cell_keys = self.clean_cell_keys(data)
+        queried_objects = self.query_database(cell_keys)
+        if queried_objects:
+            return self.prepare_location(queried_objects)
+
```
hannosch: Maybe explicitly call out the contract by doing:
```
def clean_cell_keys(self, data):
    raise NotImplementedError

def prepare_location(self, queried_objects):
    raise NotImplementedError
```

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+
+class CellLocationProvider(AbstractCellLocationProvider):
+    models = [Cell, OCIDCell]
+    log_name = 'cell'
+
+    def prepare_location(self, queried_objects):
+        self.stat_count('cell_found')
```
hannosch: cell_found can be removed

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
-            # but retain the cell countries as a likely match as well.
-            cell_countries.append(geoip['country_code'])
-
-        stats_client.incr('%s.country_from_geoip' % api_name)
-        geoip_res = {
-            'lat': geoip['latitude'],
-            'lon': geoip['longitude'],
-            'accuracy': geoip['accuracy'],
+    def prepare_location(self, queried_objects):
+        self.stat_count('cell_lac_found')
```
hannosch: cell_lac_found can go

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
-    }
+        :param items: A list of elemenets to cluster.
+        :param distance_fn: A pairwise distance_fnance function over elements.
+        :param threshold: A numeric thresholdold for clustering;
+                          clusters P, Q will be joined if
+                          distance_fn(a,b) <= threshold,
+                          for any a in P, b in Q.
 
+        :returns: A list of lists of elements, each sub-list being a cluster.
+        """
```
hannosch: In order for the doc string to be picked up, `docs/internal_api/locate.rst` needs to change from an autofunction to a autoclass (http://sphinx-doc.org/ext/autodoc.html#directive-autoclass).

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+                wifis.append(wifi)
+
+        # Estimate signal strength at -100 dBm if none is provided,
+        # which is worse than the 99th percentile of wifi dBms we
+        # see in practice (-98).
+
+        wifi_signals = dict([(w['key'], w['signal'] or -100) for w in wifis])
+        wifi_keys = set(wifi_signals.keys())
+
+        return wifis, wifi_signals, wifi_keys
```
hannosch: I prefer brackets around tuples, making them more explicit.

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+
+    def query_database(self, wifi_keys):
+        queried_wifis = []
+        try:
```
hannosch: I'd put the check here:
```
if len(wifi_keys) < MIN_WIFIS_IN_QUERY:
    return []

queried_wifis = []
try:
```
At this point it won't screw up the stats, and prevents us from doing an accidental query for all wifi keys. As a plus it avoids doing a db query when we only have a single wifi and won't return a result based on it, no matter whether we find the one or not.

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
-                lat=cell.lat,
-                lon=cell.lon,
-                range=cell.range)
-            if type(cell) is CellArea:
-                validated['cell_lac_network'].append(network)
+            if len(clusters) == 0:
+                self.stat_count('wifi.found_no_cluster')
             else:
-                validated['cell_network'].append(network)
+                self.stat_count('wifi_found')
```
hannosch: wifi_found can go

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+                    'lat': geoip['latitude'],
+                    'lon': geoip['longitude'],
+                    'accuracy': geoip['accuracy'],
+                }
+
+        if not (location or country):
+            self.stat_count('no_country')
+            self.stat_count('no_geoip_found')
+
+        return location, country
```
hannosch: brackets around tuples please :)

ichnaea/service/locate.py
```diff
@@ -122,496 +100,536 @@ def map_data(data):
...
+            location = geoip_location
+            location_provider_used = self.geoip_provider
+
+        if not location:
+            self.stat_count('miss')
+
+        for location_provider in\
+                self.location_providers + [self.geoip_provider]:
+            if location_provider is location_provider_used:
+                location_provider.log_used()
```
hannosch: This isn't quite what we want. For each request we want to have exactly one used/unused metric. Let's say we have a request in which the user provides both wifi and cell data. In this case we want to have either a wifi_hit or a wifi_miss metric and no other metrics, for wifi the caveat being that a user sending a single wifi data point should be treated the same as if the user didn't send any wifi data all.

If the user only sends us cell data, we want to have either a cell_hit or cell_miss metric, independent on whether we answer the query based on cell/cell_area/ocid_cell/ocid_cell_area data. The metric is "did we answer a cell based query with a cell based answer".

hannosch: Thinking about this, maybe each provider can return an additional three-state value, for example: True == provider was used and hit, False == provider was used and missed, None == provider wasn't used at all. The first provider which returns a `is not None` result can then update the location_provider_used variable to (class, True/False) value. The following providers won't update the location_provider_used variable anymore, as it's not None anymore.

PR #413, opened 2015-03-04
Add contribute.json file.

Osmose:
Add a contribute.json file a la http://contribute.paas.allizom.org/.

### DIFF
--- a/contribute.json
+++ b/contribute.json
@@ -1,0 +1,26 @@
+{
+    "name": "Ichnea",
+    "description": "Mozilla Ichnaea is an application to provide geo-location coordinates from other sources of data (cell or WiFi networks, GeoIP, etc.).",
+    "repository": {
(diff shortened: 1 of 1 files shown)


hannosch:
@Osmose is there a list of commonly used keywords somewhere? Or a list of keywords used by other projects? I've seen http://contribute.paas.allizom.org/examples but that's a very small list of examples, is there some other place where projects using this standard are listed?

Osmose:
That's the only list, and there's no real standard for keywords (there's not really enough projects using it yet for there to be an informal one).

I guess you should just think of the use cases of keywords (like finding projects based on programming language or by topic of what they cover) and try to put things useful to those in. I left it intentionally short so projects can figure out their own keywords.

hannosch:
Ok, wasn't sure if this was related to keywords used on for example mozillians.org.

Issue #351, opened 2014-11-21
Use new ocid_cell_area in location look-ups


hannosch:
After #348 and #349 are done, we can use the new ocid_cell_area data in the service.locate logic. Having #349 done should help make the logic simpler, as both area data have the same data format and column names.

This touches the code and tests for `ichnaea.service.locate.query_cell_networks` and `search_cell` / `search_cell_lac`. We generally prefer using our own data if we have any over the ocid data. But we can now get into a couple of new situations, we need to take care of. We can get multiple cells from multiple location area in the requests. If either we or ocid has an exact cell match, we should use it. If we get at least one exact cell match, we should also use cells from the same source (us vs. ocid) from the same area to do potential trilateration. We don't mix data from the two sources, as the data points might disagree a lot. We also don't use data from multiple location areas at the same time. A common scenario here is getting data from a multi-SIM device and so getting data for a GSM and UMTS cell at the same time, where we don't want to do trilateration across these cells.

As an aside, it might be easier to merge the search_cell/search_lac_cell functions into one, query all the relevant cells inside that one function and keep the individual cells and areas separate. The current code works, as both cells and areas come from the same model and have exactly the same structure.

There are a whole bunch of test cases to write for this. Combinations I can think of:
* No cells
* One radio type:
  * One cell in one area
  * Three cells in one area
  * Multiple cells from multiple areas
    * One area with one cell, a different area with two cells
    * Two areas with two cells each
* Two radio types:
  * One cell from one area / radio each
  * One cell from one area / radio, two cells from a different area / radio
  * Two cells from each area / radio

On the other axis we have:
* Neither mls nor ocid has any of the data
* MLS has data about cell and area
* MLS has no data, ocid has data about cell
* MLS has no data, ocid has data about area
* MLS has data about area, ocid has data about cell and area
* MLS has data about one out of multiple cells, OCID has data about more than one of the cells

We don't need to test all of these combinations, but just the "interesting" ones.

I think as an algorithm we want to extend the current approach to group things not just by location area, but by source / area combination. Than pick the source / location area combination with the most found cells in it, breaking a tie by first preferring mls data over ocid and than using the area with the smaller overall area_range.

After this we should end up with zero or one area record and zero, one or multiple cells in that area. We can then either use the area, the one cell or do trilateration over the multiple cells.

Once ocid has finished their data cleanup (sometime later this / maybe next year), we can look at using a different approach, like looking at the number of samples each record is based on to judge quality. For now we trust mls over ocid.

PR #414, opened 2015-03-05
Use new ocid_cell_area in location look-ups #351

jaredkerim:
Here is a first pass on #351 for review.  This uses the existing framework of provider types with different priorities to create new instances of providers which use the OCIDCell and OCIDCellArea models to perform location lookups.

I moved the tests around a lot, first to split things up by what they're testing (cell, geoip, wifi, country) and then within Cell I moved most of the tests to a Cell base class which is model agnostic, and then created subclasses for each pair of (Cell, CellArea) * (OCIDCell, OCIDCellArea).

For the cases where we use MLS Cell and OCID CellArea, we will prefer the MLS Cells over the OCID Cell Areas according to the current implementation which disagrees with some tests that suggest we should prefer LACs over Cells in some situations.   These tests have been overridden in that subclass to show the expected behaviour.

I kept this pull request unsquashed so it would be clearer which steps I took, in what order.

Please take a look @hannosch if you can and leave some feedback.  I'm not sure that we're testing every case you suggested, and I'm not sure whether all the behaviours of this implementation match what you described in your original writeup for the ticket.

So let me know where the gaps are and I can take a look.


ichnaea/service/locate.py
```diff
@@ -378,22 +403,35 @@ def locate(self, data):
...
     """
-    models = (Cell, OCIDCell)
+    models = (Cell, )
+
+
+class OCIDCellLocationProvider(AbstractCellLocationProvider):
+
+    """
+    A CellLocationProvider implements a cell location search using
+    the Cell and OCID Cell models.
```
hannosch: Docstring is outdated, it's either Cell or OCID now :)

ichnaea/service/locate.py
```diff
@@ -97,14 +107,14 @@ def map_data(data, client_addr=None):
     return mapped
 
 
-class AbstractResult(object):
-    """The result of a location provider query."""
+class AbstractLocation(object):
+
```
hannosch: What pep8 rule is this? My pep8/flake8 doesn't warn me about this one. My flake8 config has "ignore = E226,N802,N806,N811", with the pep8-naming plugin installed.

ichnaea/service/locate.py
```diff
@@ -97,14 +107,14 @@ def map_data(data, client_addr=None):
     return mapped
 
 
-class AbstractResult(object):
-    """The result of a location provider query."""
+class AbstractLocation(object):
```
hannosch: I agree result wasn't the best of names here. I wrote Location first, but that leads to "PositionLocation", which really is the same concept twice and looked funny to me. I'm also not sure describing a country code as a "country location" matched that well. We might also get timezone as a third type, to we'd have a timezone location, which really is just a UTC offset or timezone code. Nitpicking, naming is hard, I don't have a better suggestion.

ichnaea/service/locate.py
```diff
@@ -36,6 +38,14 @@
 # helper class used in searching
 Network = namedtuple('Network', ['key', 'lat', 'lon', 'range'])
 
+# Data sources for location information
+
+
+class DataSource(IntEnum):
+    MLS = 1
```
hannosch: Maybe naming this "OWN" or "OURS" would could avoid the MLS reference here. In theory ichnaea is a general purpose software and MLS just one instance of it running ... but then again the website content is full of MLS references. Nothing to do here, just a thought.

ichnaea/service/locate.py
```diff
@@ -378,22 +403,35 @@ def locate(self, data):
 
 
 class CellLocationProvider(AbstractCellLocationProvider):
+
     """
     A CellLocationProvider implements a cell location search using
     the Cell and OCID Cell models.
     """
```
hannosch: Docstring outdated, only single model is used. Maybe we could change `models = (...)` to `model = ...` now that we use a single model everywhere.

ichnaea/service/locate.py
```diff
@@ -706,80 +762,85 @@ def __init__(self, db_sources, *args, **kwargs):
...
+                    best_location):
                 # If this location is more accurate than our previous one,
                 # we'll use it.
-                result = provider_result
+                best_location = provider_location
+                best_location_provider = provider
 
-            if result.accurate_enough():
-                # Stop the loop, if we have a good quality result.
+            if best_location.accurate_enough():
```
hannosch: Are we missing a test here where the first provider doesn't return a result at all? Shouldn't best_location still be None then here, so `best_location.accurate_enough()` would fail? I kinda liked the empty result idea, which avoided this sort of "is None" checking. And spelling that out as a concrete class, like you did in between looked pretty good to me.

hannosch: Ah no, correcting myself. Even if the first provider doesn't find a result it still returns a PositionLocation object, just maybe with no data in it. So by the time we hit this line, we always have an object.

hannosch:
Let's talk about the tests this evening. Splitting them up helps, but the cell stuff isn't quite right. We shouldn't really test for database inconsistencies in the common case, so if we have a cell model, we should also always have a cell area model. Same for ocid, where we should always have a pair of those two. In some cases our query may not match the individual cell and only hit the area, but we should always create them in pairs. That means we should remove a test like `test_cell_disagrees_with_lac` which uses a cell / ocid_area combination.

I think we should also make it clearer which tests are about the internals of one provider, vs. which ones are about the combination of multiple providers.

And as a crazy thought/brainstorming: Maybe we could split up the locate.py module into separate chunks. The packages inside `ichnaea.service` are all about public HTTP endpoints, so I'd maybe create a new top-level structure like:

ichnaea.locate.base (abstract location, statslogger, abstract provider)
ichnaea.locate.cell (providers for both country/position)
ichnaea.locate.country (country location, country searcher)
ichnaea.locate.geoip
ichnaea.locate.wifi
ichnaea.locate.position (position location, position searcher)
ichnaea.locate.searcher (abstract location searcher)

Maybe some of those modules would be too small. If we had such a structure, we could add tests with one test module per code module (ichnaea.locate.tests.test_base, ichnaea.locate.tests.test_cell, ...). At that point it may even be easier to just write the tests from scratch, so we cover each code line with at least one new test and completely remove the old tests, as they really don't match the new code structure at all anymore.

hannosch:
And maybe to limit the scope of this bug, we could take all but the big test re-arrangement (b0346222f480a5dca0cb3a0cd24fa033d194c191) and merge those. And then go back and either you or me can work separately on the code split-up / test refactoring.

PR #424, opened 2015-04-21
Update raven to latest version, requires a Sentry 7 server.

hannosch:
This manages the transport manually, using gevent for webapp, threaded for celery and sync for scripts and tests.

We also pull in certifi, as it's automatically used internally by raven when using a https transport.

This is waiting on ops changes.


hannosch:
Merged via squashed 405ad34df5577e977faa4ccbf360aad808616f39.

Issue #425, opened 2015-04-22
Refactor API Key Logic


jaredkerim:
- Query the API Key object in the check_api_key decorator
- Pass that object to the decorated view
- If no API Key object can be found, pass in an empty object

hannosch:
Committed with some small adjustments, I kept the raven/stats_client locals to avoid going over the 80 chars line length.

Using the HashKey mixins for the api key model shortened the query call and made it possible to put the "exclude email/description" logic onto the model without having to put that everywhere. I'd rather avoid loading those two, as its 500 bytes we usually don't need, and not loading them in the first place also avoids leaking them in sentry reports, without us having to scrub the email from the locals later.

PR #426, opened 2015-04-22
Refactor API Key Logic #425

jaredkerim:
-Query the API Key object in the check_api_key decorator
-Pass that object to the decorated view
-If no API Key object can be found, pass in an empty object


jaredkerim:
@hannosch r?

Issue #421, opened 2015-03-26
Implement fallback location provider


hannosch:
We should add support for querying a secondary fallback location provider, if we aren't satisfied with our own location result.

This includes:
- Add new settings to the config file, maybe fallback_url, fallback_apikey, fallback_ratelimit, fallback_cache_timeout (could make this into a separate section in the ini file, but the code reading the file makes that a bit harder)
- Add new column to the api_key table, a boolean fallback_allowed
- Add a new location provider doing a "geolocate" type query to the fallback provider if the current api key is allowed to do so
- Add rate limitation logic, only allow <fallback_ratelimit config option> number of requests, potentially reusing / extending the code in ichnaea.service.base.rate_limit. The config option might have to be a simple expression like "5/5s" stating "allow 5 requests per 5 second interval", only supporting seconds as the time interval. The existing rate_limit function disables rate limitation if redis is not available. For outbound requests we want to do the opposite and not do a request if redis is unavailable.
- In the first round, only do a fallback request if we couldn't get any non-geoip data from our own database. Also never do a fallback request if the request contained no valid cell/wifi identifiers (this includes only getting a single wifi data point). If all we could get is a geoip result, there's no value in doing a fallback.
- In the fallback query, use the custom syntax and add`"fallbacks": {"lacf": 1}` to the geolocate query (see http://combain.com/api/ fallbacks or the WCDMA lacf example. This should match what we are doing.
- Add caching of the responses we got from the fallback (in Redis). Use the fallback_cache_timeout config setting to determine the cache timeout value. The config option can be specified as an integer in seconds. At first only do response caching if we get a query containing a single cell network. The cache key can then consist of some cache prefix and the cell hash key. As a value we can store the actual JSON response (lat/lon/accuracy) or some token to cache not found responses.

Still undefined:
- What exact metrics should we gather? Something about how many fallback requests we make, how many times do we reach the rate limit. How does the fallback tie into our KPI metrics (the provider group stuff), as we cannot reliably distinguish between getting a cell or wifi based response from the fallback. We also want some metrics on cache effectiveness / hit rates.
- Do we want to do caching for Wifi based queries? A cache key consisting of all bssid's could get really large. For queries based on multiple data points the signal strength data also contributes significantly to a better position estimate, so we might need to make that part of the cache key as well. Might lead to cache key explosion without any cache hits. Probably defer to later.

hannosch:
What remains of this are:
[...]

PR #429, opened 2015-04-24
Add Fallback Provider #421

jaredkerim:
- Add FallbackProvider
- Add allow_fallback to ApiKey
- Add fallback parameters to ichnaea.ini
- Add should_locate method to provider/searcher contract
- Add tests

### DIFF
--- a/ichnaea/locate/provider.py
+++ b/ichnaea/locate/provider.py
@@ -4,6 +4,7 @@ from functools import partial
 import operator
 
 import mobile_codes
+import requests
 from sqlalchemy.orm import load_only
 
 from ichnaea.constants import (
@@ -41,6 +42,7 @@ class DataSource(IntEnum):
     Internal = 1
     OCID = 2
     GeoIP = 3
+    Fallback = 4
 
 
 class Provider(StatsLogger):
@@ -57,14 +59,24 @@ class Provider(StatsLogger):
     location_type = None
     source = DataSource.Internal
 
-    def __init__(self, session_db, geoip_db, *args, **kwargs):
+    def __init__(self, session_db, geoip_db, settings, *args, **kwargs):
         self.session_db = session_db
         self.geoip_db = geoip_db
+        self.settings = settings
         self.location_type = partial(self.location_type, source=self.source)
         super(Provider, self).__init__(*args, **kwargs)
 
+    def should_locate(self, data, location):
+        """
+        Given location query data and a possible location
+        found by another provider, check if this provider should
+        attempt to perform a location search.
+        """
+        return True
+
     def locate(self, data):  # pragma: no cover
-        """Provide a location given the provided query data (dict).
+        """
+        Provide a location given the provided query data (dict).
 
         :rtype: :class:`~ichnaea.locate.location.Location`
         """
@@ -531,3 +543,79 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
 
 class GeoIPCountryProvider(BaseGeoIPProvider):
     location_type = Country
+
+
+class FallbackProvider(Provider):
+    log_name = 'fallback'
+    location_type = Position
+    source = DataSource.Fallback
+
+    def _prepare_cell(self, cell):
+        return {
+            'radioType': cell.get('radio', None),
+            'cellId': cell.get('cid', None),
+            'locationAreaCode': cell.get('lac', None),
+            'mobileCountryCode': cell.get('mcc', None),
+            'mobileNetworkCode': cell.get('mnc', None),
+        }
+
+    def _prepare_wifi(self, wifi):
+        return {
+            'macAddress': wifi.get('key', ''),
+        }
+
+    def _prepare_data(self, data):
+        cell_queries = [
+            self._prepare_cell(cell) for cell in data.get('cell', [])]
+        wifi_queries = [
+            self._prepare_wifi(wifi) for wifi in data.get('wifi', [])]
+
+        return {
+            'cellTowers': cell_queries,
+            'wifiAccessPoints': wifi_queries,
+        }
+
+    def should_locate(self, data, location):
+        empty_location = not location.found()
+        weak_location = location.source >= DataSource.GeoIP
+        cell_found = data.get('cell', [])
+        wifi_found = len(data.get('wifi', [])) > 1
+        return (
+            self.api_key.allow_fallback and
+            (empty_location or weak_location) and
+            (cell_found or wifi_found)
+        )
+
+    def locate(self, data):
+        location = self.location_type(query_data=False)
+
+        query_data = self._prepare_data(data)
+
+        try:
+            response = requests.post(
+                self.settings['url'],
+                headers={'User-Agent': 'ichnaea'},
+                json=query_data,
+                timeout=5.0,
+                verify=False,
+            )
+            response.raise_for_status()
+        except requests.exceptions.RequestException:
+            self.raven_client.captureException()
+            response = None
+
+        if response:
+            try:
+                json_data = response.json()
+                lat = json_data['location']['lat']
+                lon = json_data['location']['lng']
+                accuracy = json_data['accuracy']
+                location = self.location_type(
+                    lat=lat,
+                    lon=lon,
+                    accuracy=accuracy,
+                )
+            except (KeyError, TypeError):
+                self.raven_client.captureException()
+
+        return location
--- a/ichnaea/locate/searcher.py
+++ b/ichnaea/locate/searcher.py
@@ -5,6 +5,7 @@ from ichnaea.locate.provider import (
     CellAreaPositionProvider,
     CellCountryProvider,
     CellPositionProvider,
+    FallbackProvider,
     GeoIPCountryProvider,
     GeoIPPositionProvider,
     OCIDCellAreaPositionProvider,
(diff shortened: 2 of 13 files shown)


ichnaea/locate/provider.py
```diff
@@ -531,3 +545,63 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
...
+
+    def locate(self, data):
+        location = self.location_type(query_data=False)
+
+        cell_queries = [{
+            'cellId': cell.get('cid', None),
+            'locationAreaCode': cell.get('lac', None),
+            'mobileCountryCode': cell.get('mcc', None),
+            'mobileNetworkCode': cell.get('mnc', None),
+        } for cell in data.get('cell', [])]
```
hannosch: This is missing the radioType field.

ichnaea/locate/provider.py
```diff
@@ -531,3 +545,63 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
...
+        wifi_queries = [{
+            'macAddress': re.sub(
+                '(..(?!$))', '\g<0>:', wifi.get('key', None)),
```
hannosch: Do we need to add these colons? Or does it also work without adding them?

ichnaea/locate/provider.py
hannosch: For radio, make sure we send a plain text string, and not one of our enum's. We only need to send one radioType field in each cell entry, nothing at the top level.

ichnaea/locate/provider.py
```diff
@@ -531,3 +545,63 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
...
+        try:
+            response = requests.post(
+                self.fallback_url, json=query_data)
+        except requests.exceptions.RequestException:
+            response = None
+
+        if response:
+            try:
+                location_data = json.loads(response.content)
+            except ValueError:
```
hannosch: Why not just use `response.json` - via http://docs.python-requests.org/en/latest/user/quickstart/#json-response-content

ichnaea/locate/provider.py
```diff
@@ -531,3 +545,63 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
...
+                location_data = {}
+
+            location = self.location_type(
+                lat=location_data.get('lat', None),
+                lon=location_data.get('lon', None),
+                accuracy=location_data.get('accuracy', None),
+            )
```
hannosch: This looks like the wrong format to me, according to the docs we should expect a nested structure like:
```
{
	"location": {
		"lat": 59.331706,
		"lng": 18.079073
	},
	"accuracy": 500
}
```

ichnaea/locate/searcher.py
```diff
@@ -118,6 +120,9 @@ class PositionSearcher(Searcher):
         ('wifi', (
             WifiPositionProvider,
         )),
+        ('fallback', (
+            FallbackProvider,
+        )),
```
hannosch: Could we reorder this list and put the geoip provider last? Would that avoid the "weak location" check inside the fallback's should_locate method?

ichnaea/locate/provider.py
```diff
@@ -531,3 +545,63 @@ class GeoIPPositionProvider(BaseGeoIPProvider):
...
+
+        query_data = {
+            'cellTowers': cell_queries,
+            'wifiAccessPoints': wifi_queries,
+        }
+
+        try:
+            response = requests.post(
+                self.fallback_url, json=query_data)
+        except requests.exceptions.RequestException:
```
hannosch: Can you add an override for the user-agent header? So it becomes something like I did in the export code at https://github.com/mozilla/ichnaea/blob/master/ichnaea/data/export.py#L105

Something like user-agent: ichnaea, so we don't leak what specific version of urllib/python we are using internally.

I'd also add the other two things I did there: 1. add a timeout, for this lookup make it 5 seconds and disable ssl verification for now.

IIRC we have to do a bunch of work to get SNI support working in Python 2.6.

ichnaea/locate/tests/test_provider.py
```diff
@@ -505,3 +512,146 @@ def test_geoip_country(self):
...
+    TestProvider = FallbackProvider
+
+    def setUp(self):
+        super(TestFallbackProvider, self).setUp()
+
+        self.provider.api_key.allow_fallback = True
+
+        requests_patcher = mock.patch('ichnaea.locate.provider.requests.post')
+        self.mock_requests = requests_patcher.start()
+        self.addCleanup(requests_patcher.stop)
```
hannosch: For requests mocking, we already use the extra helper library mock_requests (https://requests-mock.readthedocs.org/en/latest/). There's a couple examples over here: https://github.com/mozilla/ichnaea/blob/master/ichnaea/data/tests/test_export.py#L139

ichnaea/locate/tests/test_provider.py
```diff
@@ -505,3 +512,146 @@ def test_geoip_country(self):
...
+        self.mock_requests.return_value = response
+
+        cells = [{
+            'cid': cid,
+            'radio': Radio.gsm.name,
+            'mcc': GB_MCC,
+            'mnc': 1,
+            'lac': 1,
+        } for cid in range(1, 3)]
+        wifis = [{'key': '001122334455'}, {'key': '112233445566'}]
```
hannosch: Can you use our factory boy test factories to get these cell/wifi keys? Someone told me to use them :))

ichnaea/locate/tests/test_provider.py
```diff
@@ -505,3 +512,146 @@ def test_geoip_country(self):
...
+        self.mock_read_config().get_map.return_value = {
+            'fallback_url': self.fallback_url,
+        }
+
+    def test_successful_call_returns_location(self):
+        response_location = {
+            'lat': 1.0,
+            'lon': 1.0,
+            'accuracy': 100,
+        }
```
hannosch: Noticed this in the actual logic, this is the wrong response format. This looks like our search response, but not the geolocate API response.

hannosch:
One more thing: I'd change the way we specify the config settings a bit. As part of my export work, I've spent a bit of time to clean that area of the code up, so we can specify multiple config sections now. So I'd change the config file over to have something like:
```
[locate:fallback]
url = https://example.com/?key=secret
ratelimit = 
cache_timeout = 
```
Specifying the api_key separately from the url isn't needed, and your code expects it in one setting as well.

To provide the test default values, you can then set up a new section in https://github.com/mozilla/ichnaea/blob/master/ichnaea/tests/base.py#L60

These config settings get parsed as part of the startup process, over here: https://github.com/mozilla/ichnaea/blob/master/ichnaea/webapp/config.py#L28. That leads to them being available as part of the request, as `request.registry.settings`. So something like `request.registry.settings.get('locate:fallback', {}).get('url', None)` will get you the url.

I'd pass the entire settings into the searcher. As part of the provider setup (https://github.com/mozilla/ichnaea/blob/master/ichnaea/locate/searcher.py#L33), I'd let it pull out the appropriate config section (`settings.get('locate:' + provider_group, {}`) out of the general settings and pass that into each provider.

That way the fallback provider ends up taking one more settings argument, which is a plain dict which at first is just a `{'url': 'https://example.com/?key=secret'}`, while later that dict automatically gets the extra settings.

Inside the tests we can avoid mocking this setting and just explicitly pass it in. Inside the application code, we don't need to call read_config and parse the config file on each request. Should any of the other providers need config settings in the future, it's trivial to add a new config section and it magically ends up in the right place. We may want to move some of the clustering constants into such settings in the future.

hannosch:
I started looking into what kind of metrics we want to capture here, and came up with #430 first to simplify this. Once that is done we can revisit adding new metrics for the fallback provider, probably something like a timer called `locate.fallback.lookup` around the outgoing request, and counters to track the HTTP response status codes under `locate.fallback.lookup_status.200`, `locate.fallback.lookup_status.400`, ...

Not quite sure yet what kind of quality metrics we want (for example fallback returned wifi / cell / no result - how would we determine this ... or just a histogram of returned accuracy?).

ichnaea/locate/searcher.py
jaredkerim: Hrmm.... this needs some further research.  I'll look at it tomorrow.

jaredkerim:
@hannosch r? should be good to merge

ichnaea/locate/searcher.py
hannosch: Ok, then ignore this. The weak check is actually a sensible one. We'll probably end up changing this bit soon enough. Maybe something like "if fallback is allowed for this request, do that first", so it's going to be more of a preferred provider, or do both lookups in our own data and the fallback provider and collect some metrics around how those two results differ from each other. Not sure yet, so leave as is for now :)

jaredkerim:
There's two methods, ._to_json() and ._to_json_value()  The one you're thinking of is ._to_json(), ._to_json_value() returns just the dict we want.  I agree it's a long ugly method call, what would be nice is a different set of serializers for the same models that serialize them to the different formats depending on which API we're talking to.  We should look into some serialization modules, maybe sqlalchemy has something for serializing models to json, but we should probably do that in a separate ticket.

jaredkerim:
That's odd, why didn't your comments get associated with lines?  It was kind of a puzzle trying to figure out which lines you were talking about!  Anyways I think that's everything, r? @hannosch again?

hannosch:
@jaredkerim I'll have a look on Monday. As for line numbers, my guess is force-pushing destroys those, as the comments are only attached to lines from the old commit hash. I still prefer to never force push during a review cycle, and only merge and rebase at the very end.

hannosch:
And as for the `_to_json_value` method, you are right, I looked at the wrong one. But even `_to_json_value` has a problem for cell keys: It encodes the radio field as an int, and not as `.name`, so it comes out as `{'radio': 0, mcc: 123, ...}` which is want the query wants either.

To be fair in the test where you use it, it pretty much doesn't matter, as the entire logic only checks for the number of items in the cell and wifi lists, and you could put pretty much anything in there.

PR #432, opened 2015-05-05
Add Rate Limiting to Fallback Provider #421

jaredkerim:
- Pass redis client to providers
- Add rate limiting settings
- Check for rate limiting in FallbackProvider
- Add a rate limiting test
- Clear redis between tests


ichnaea/locate/tests/test_provider.py
```diff
@@ -53,9 +54,11 @@ class TestProvider(Provider):
     def setUpClass(cls):
         DBTestCase.setUpClass()
         GeoIPIsolation.setup_geoip(raven_client=cls.raven_client)
+        RedisIsolation.setup_redis()
 
     @classmethod
     def tearDownClass(cls):
+        RedisIsolation.teardown_redis()
         GeoIPIsolation.teardown_geoip()
         DBTestCase.tearDownClass()
```
hannosch: Did you solve the redis isolation problem? If not, you might want to add this bit, so redis gets cleaned after each test method:

```
def tearDown(self):
    self.cleanup_redis()
    self.teardown_session()
```
That's what AppTestCase does, which is almost the same combination of base classes, except that it also sets up a wsgi app.

ichnaea/rate_limit.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def rate_limit(redis_client, key, maxreq=0, expire=86400):
+    if maxreq:
+        try:
+            redis_client.set(key, 0, ex=expire, nx=True)
+            count = redis_client.incr(key, 1)
+            return count > maxreq
+        except ConnectionError:  # pragma: no cover
+            # If we cannot connect to Redis, disable rate limitation.
```
hannosch: This behavior is fine for users calling us. But when we call an external party, we don't want to make any requests if we cannot figure out the current rate limit.

So this needs one more function argument, something like `fail/pass_on_error`.

ichnaea/locate/searcher.py
```diff
@@ -27,7 +27,8 @@ class Searcher(StatsLogger):
 
     provider_classes = ()
 
-    def __init__(self, session_db, geoip_db, settings, *args, **kwargs):
+    def __init__(self, session_db, geoip_db, settings,
+                 redis_client=None, *args, **kwargs):
```
hannosch: I'd move redis_client before the settings argument and make it required. It's more similar to session and geoip as some sort of external connection object.

ichnaea/locate/provider.py
```diff
@@ -59,10 +61,12 @@ class Provider(StatsLogger):
     location_type = None
     source = DataSource.Internal
 
-    def __init__(self, session_db, geoip_db, settings, *args, **kwargs):
+    def __init__(self, session_db, geoip_db, settings,
+                 redis_client=None, *args, **kwargs):
```
hannosch: Same argument order here, move redis_client before settings (see searcher comment).

ichnaea/locate/provider.py
```diff
@@ -596,6 +600,17 @@ def should_locate(self, data, location):
     def locate(self, data):
         location = self.location_type(query_data=False)
 
+        if self.redis_client:
```
hannosch: if redis_client becomes a required argument, the condition here can check the ratelimit setting instead.

ichnaea/locate/tests/test_searcher.py
```diff
@@ -12,7 +12,7 @@
 )
 
 
-class SearcherTest(DBTestCase, GeoIPIsolation):
+class SearcherTest(GeoIPIsolation, DBTestCase):
```
hannosch: Any specific reason to reorder these? I think I stuck to putting the testcase subclass as the first one in all other places. The *Isolation classes never implement any of the setUp/tearDown/class methods, but I kept the testcase as the first one to make sure none of them sneak in and make the super calls even more complicated.

ichnaea/tests/base.py
```diff
@@ -76,6 +76,7 @@
     },
     'locate:fallback': {
         'url': 'http://127.0.0.1:9/?api',
+        'ratelimit': 10,
```
hannosch: Can you make this a string here? At this point there's no magic yet which would have converted it from a string to an int.

ichnaea/locate/provider.py
```diff
@@ -596,6 +600,17 @@ def should_locate(self, data, location):
...
+            rate_key = 'fallback:{time}'.format(time=int(time.time()))
+            limit_reached = rate_limit(
+                self.redis_client,
+                rate_key,
+                maxreq=self.settings['ratelimit'],
```
hannosch: From the config file the setting comes in as a string. Either we need to cast it to an int here or provide a custom `__init__` for the fallback provider which does the parsing and assigns it as something like `self.ratelimit`. I'd prefer to put the parsing logic into an init.

ichnaea/locate/provider.py
```diff
@@ -596,6 +600,17 @@ def should_locate(self, data, location):
...
+                expire=1,
```
hannosch: I'd set this to a bit more than 1 second, maybe 60 seconds. Since the timestamp is client side generated, it's susceptible to clock-skew (dean tells me we are usually seeing a couple seconds, outliers at 10-20 seconds of skew on machines). I wouldn't worry about that too much for this use-case, but keeping the keys around for a bit longer shouldn't hurt. A client that is a couple second behind the others, would still try to increment one of the existing older keys, and get throttled by that. If the keys are gone, we could have each of the three servers essentially write and increment its own counter without ever overlapping with the other counters.

ichnaea/locate/tests/test_provider.py
```diff
@@ -525,6 +529,7 @@ def setUp(self):
         self.provider.api_key.allow_fallback = True
         self.provider.settings = {
             'url': 'http://127.0.0.1:9/?api',
+            'ratelimit': 5,
```
hannosch: Make this a string here as well.

ichnaea/locate/tests/test_provider.py
```diff
@@ -700,3 +705,32 @@ def test_should_not_provide_location_if_non_geoip_location_found(self):
...
+                })
+
+                self.assertTrue(location.found())
+
+            location = self.provider.locate({
+                'cell': self.cells,
+                'wifi': self.wifis,
+            })
+
+            self.assertFalse(location.found())
```
hannosch: This assumes this call is happening in the same second, as the successful attempts earlier. If we are unlucky, the second will roll over right in between these calls.

I'd rather make this a bit more stable:
* Have one test that checks it can do queries up until incl. the rate limit.
* Add a second test that manually constructs and adds the rate limit exceeded keys for the next three seconds into Redis, than calls the provider and checks that it respected the limit. By adding the keys for a couple continuous seconds, we should be safe from second roll-over in the middle of it.

ichnaea/rate_limit.py
```diff
@@ -0,0 +1,12 @@
+from redis import ConnectionError
+
+
+def rate_limit(redis_client, key, maxreq=0, expire=86400):
+    if maxreq:
+        try:
+            redis_client.set(key, 0, ex=expire, nx=True)
+            count = redis_client.incr(key, 1)
+            return count > maxreq
```
hannosch: I liked the former version with a pipeline better, that only made one network request instead of two. Calling `incr` on a non-existent key will just set it to 1, so the explicit "set to zero" isn't required.

ichnaea/tests/base.py
```diff
@@ -299,6 +300,10 @@ def cleanup_tables(cls, engine):
 
 class RedisIsolation(object):
 
+    def tearDown(self):
+        self.redis_client.flushdb()
+        super(RedisIsolation, self).tearDown()
+
```
hannosch: Ah, I decided against putting any of the standard setUp/tearDown methods on the *Isolation classes, as reasoning about what would get called with multiple inheritance gets really tricky. That's why anyone who subclasses from any of the isolation classes has to manually call the setup_/teardown_/cleanup methods provided on those isolation classes.

I'd rather keep it as it was. I mentioned the only required change at https://github.com/mozilla/ichnaea/pull/432/files#r29719552

ichnaea/tests/test_rate_limit.py
```diff
@@ -0,0 +1,57 @@
...
+                api_key,
+                maxreq=maxreq,
+                expire=expire,
+            ))
+        self.assertTrue(rate_limit(
+            self.redis_client,
+            api_key,
+            maxreq=maxreq,
+            expire=expire,
+        ))
```
hannosch: I know this is just a moved existing test, but if you feel like it, https://github.com/mozilla/ichnaea/pull/432/files#r29757957 applies here as well (second rollover right before the assertTrue call).

hannosch:
I'd leave the testcase/super changes out of this one. If we want to go there, we can do it separately from the rest - and I still have to watch Raymond's super talk :)

ichnaea/rate_limit.py
jaredkerim: Oh that's so weird, I tried doing that and it was raising an error, but I just tried again and it's magically not, maybe I was doing something differently.  Yes your'e right.

ichnaea/tests/base.py
jaredkerim: I don't understand your reasoning here.  If you include the RedisIsolation mixin in a test case, you want to be able to access the redis client inside your tests.  If you have tests which might interact with a redis database, you want to flush that database after each test is run.  Where should the code for flushing the db live?

jaredkerim: Oh sorry I see what you mean, you want each subclassing test case to explicitly have to call that out.  I don't like that behaviour, but since it's the existing pattern in the project I'll make that change.

jaredkerim:
Feedback implemented r? @hannosch

ichnaea/locate/tests/test_provider.py
```diff
@@ -65,11 +71,16 @@ def setUp(self):
         self.provider = self.TestProvider(
             session_db=self.session,
             geoip_db=self.geoip_db,
-            settings={},
+            settings=self.settings,
+            redis_client=self.redis_client,
```
hannosch: Wrong order for settings/redis_client

ichnaea/service/country/views.py
```diff
@@ -32,6 +32,7 @@ def country_view(request):
         session_db=request.db_ro_session,
         geoip_db=request.registry.geoip_db,
         settings=request.registry.settings,
+        redis_client=request.registry.redis_client,
```
hannosch: Wrong order for settings/redis_client

ichnaea/service/geolocate/views.py
```diff
@@ -44,6 +44,7 @@ def geolocate_view(request, api_key):
         session_db=request.db_ro_session,
         geoip_db=request.registry.geoip_db,
         settings=request.registry.settings,
+        redis_client=request.registry.redis_client,
```
hannosch: order

ichnaea/service/search/views.py
```diff
@@ -57,6 +57,7 @@ def search_view(request, api_key):
         session_db=request.db_ro_session,
         geoip_db=request.registry.geoip_db,
         settings=request.registry.settings,
+        redis_client=request.registry.redis_client,
```
hannosch: order

ichnaea/locate/provider.py
```diff
@@ -550,6 +554,12 @@ class FallbackProvider(Provider):
     location_type = Position
     source = DataSource.Fallback
 
+    def __init__(self, settings, *args, **kwargs):
+        self.url = settings['url']
+        self.ratelimit = int(settings['ratelimit'])
```
hannosch: can you change that to `self.ratelimit = int(settings.get('ratelimit', '0'))`, so the config is optional?

Feel free to make that into something better that uses `None` as a not-set key as well, I was too lazy ;)

hannosch:
Overall looks good, some tiny new comments. Can you do those and then squash/rebase? I'll then merge it first thing on Friday. You can start your next work based on the squashed commit.

ichnaea/locate/tests/test_provider.py
jaredkerim: Ah true but when you refer to them by name order is irrelevant.

PR #436, opened 2015-05-12
Cache single cell results for fallback #421

jaredkerim:
- When we receive only a single cell tower, cache the results
- Add tests

@hannosch r?

### DIFF
--- a/ichnaea/locate/provider.py
+++ b/ichnaea/locate/provider.py
@@ -1,11 +1,13 @@
+import simplejson as json
+import operator
+import time
 from collections import defaultdict, namedtuple
 from enum import IntEnum
 from functools import partial
-import operator
-import time
 
 import mobile_codes
 import requests
+from redis import ConnectionError
 from sqlalchemy.orm import load_only
 
 from ichnaea.constants import (
@@ -529,10 +531,13 @@ class FallbackProvider(Provider):
     log_name = 'fallback'
     location_type = Position
     source = DataSource.Fallback
+    LOCATION_NOT_FOUND = True
 
     def __init__(self, settings, *args, **kwargs):
         self.url = settings['url']
         self.ratelimit = int(settings.get('ratelimit', 0))
+        self.rate_limit_expire = int(settings.get('ratelimit_expire', 0))
+        self.cache_expire = int(settings.get('cache_expire', 0))
         super(FallbackProvider, self).__init__(
             settings=settings, *args, **kwargs)
 
@@ -587,16 +592,48 @@ class FallbackProvider(Provider):
             self.redis_client,
             self.get_ratelimit_key(),
             maxreq=self.ratelimit,
-            expire=60,
+            expire=self.rate_limit_expire,
             fail_on_error=True,
         )
 
-    def locate(self, data):
-        location = self.location_type(query_data=False)
+    def _should_cache(self, data):
+        return (
+            self.cache_expire and
+            len(data.get('cell', [])) == 1 and
+            len(data.get('wifi', [])) == 0
+        )
 
-        if self.limit_reached():
-            return location
+    def _get_cache_key(self, cell_data):
+        return 'fallback_cache_{0}'.format(sorted(cell_data.items()))
 
+    def _get_cached_result(self, data):
+        if self._should_cache(data):
+            all_cell_data = data.get('cell', [])
+            cache_key = self._get_cache_key(all_cell_data[0])
+            try:
+                cached_cell = self.redis_client.get(cache_key)
+                if cached_cell:
+                    self.stat_count('fallback.cache.hit')
+                    return json.loads(cached_cell)
+                else:
+                    self.stat_count('fallback.cache.miss')
+            except ConnectionError:
+                self.raven_client.captureException()
+
+    def _set_cached_result(self, data, result):
+        if self._should_cache(data):
+            all_cell_data = data.get('cell', [])
+            cache_key = self._get_cache_key(all_cell_data[0])
+            try:
+                self.redis_client.set(
+                    cache_key,
+                    json.dumps(result),
+                    ex=self.cache_expire,
+                )
+            except ConnectionError:
+                self.raven_client.captureException()
+
+    def _make_external_call(self, data):
         query_data = self._prepare_data(data)
 
         try:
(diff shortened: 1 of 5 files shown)


ichnaea/locate/provider.py
```diff
@@ -611,16 +614,34 @@ def limit_reached(self):
...
             maxreq=self.ratelimit,
-            expire=60,
+            expire=self.RATE_LIMIT_EXPIRE,
             fail_on_error=True,
         )
 
-    def locate(self, data):
-        location = self.location_type(query_data=False)
+    def _get_cache_key(self, cell_data):
+        return 'fallback_cache_{}'.format(sorted(cell_data.items()))
```
hannosch: That's Python 2.7 syntax, in 2.6 there needs to be a number or a name inside the curly braces.

ichnaea/locate/provider.py
```diff
@@ -1,8 +1,9 @@
+import json
```
hannosch: Can you import and use simplejson for this? In Python 2.6 the built-in json module is a slow pure-Python implementation.

ichnaea/locate/provider.py
```diff
@@ -611,16 +614,34 @@ def limit_reached(self):
...
+        return 'fallback_cache_{}'.format(sorted(cell_data.items()))
 
-        if self.limit_reached():
-            return location
+    def _fetch_cached_result(self, data):
+        all_cell_data = data.get('cell', [])
 
+        if len(all_cell_data) == 1:
+            cache_key = self._get_cache_key(all_cell_data[0])
+            cached_cell = self.redis_client.get(cache_key)
```
hannosch: Can you catch redis connection errors around this block? And if we can't connect to redis, simply return no cached value.

ichnaea/locate/provider.py
```diff
@@ -611,16 +614,34 @@ def limit_reached(self):
...
+    def _set_cached_result(self, data, result):
+        all_cell_data = data.get('cell', [])
+
+        if len(all_cell_data) == 1:
+            cache_key = self._get_cache_key(all_cell_data[0])
+            self.redis_client.set(
+                cache_key,
+                json.dumps(result),
+                ex=self.CACHE_EXPIRE,
+            )
```
hannosch: Catch redis connection errors here as well and ignore them.

ichnaea/locate/provider.py
```diff
@@ -633,24 +654,35 @@ def locate(self, data):
...
+                try:
+                    location = self.location_type(
+                        lat=location_data['location']['lat'],
+                        lon=location_data['location']['lng'],
+                        accuracy=location_data['accuracy'],
+                    )
+                except (KeyError, TypeError):
+                    self.raven_client.captureException()
+
+                self._set_cached_result(data, location_data)
```
hannosch: Wouldn't this re-set the cached value we got earlier? So basically extending the expiry time on the cached value every time? It would also prevent us from actually checking the external service again for better data.

I'd only set the value here, if it didn't come from our cache.

ichnaea/locate/provider.py
```diff
@@ -553,6 +554,8 @@ class FallbackProvider(Provider):
     log_name = 'fallback'
     location_type = Position
     source = DataSource.Fallback
+    RATE_LIMIT_EXPIRE = 60
+    CACHE_EXPIRE = (24 * 60 * 60)  # 1 day
```
hannosch: I'd prefer to get the cache expiry time from the config file. It should be possible to set the expiry to either None or 0 to disable this sort of caching. Going forward we want to make ichnaea more useful to third parties running the code, and that includes the little details.

ichnaea/locate/provider.py
```diff
@@ -611,16 +614,34 @@ def limit_reached(self):
...
+        if len(all_cell_data) == 1:
+            cache_key = self._get_cache_key(all_cell_data[0])
+            cached_cell = self.redis_client.get(cache_key)
+            if cached_cell:
+                return json.loads(cached_cell)
+
+    def _set_cached_result(self, data, result):
+        all_cell_data = data.get('cell', [])
+
+        if len(all_cell_data) == 1:
```
hannosch: We also need to check for wifi here, if we get one cell and any wifi networks, we don't want to cache, as the result is likely wifi based.

ichnaea/locate/tests/test_provider.py
```diff
@@ -768,3 +768,40 @@ def test_redis_connection_failure_prevents_external_call(self):
             })
 
             self.assertFalse(location.found())
+
+    def test_single_cell_results_cached_preventing_external_call(self):
```
hannosch: I'd add at least one test to check that we don't cache if we get one cell + one or multiple wifis. The zero/zero or two cells cases are a bit stupid and I wouldn't test for them explicitly.

ichnaea/locate/provider.py
```diff
@@ -611,16 +614,34 @@ def limit_reached(self):
...
-        if self.limit_reached():
-            return location
+    def _fetch_cached_result(self, data):
+        all_cell_data = data.get('cell', [])
 
+        if len(all_cell_data) == 1:
+            cache_key = self._get_cache_key(all_cell_data[0])
+            cached_cell = self.redis_client.get(cache_key)
+            if cached_cell:
+                return json.loads(cached_cell)
```
hannosch: Can you add one more counter to this, something like:
`self.stat_count('fallback.cache.hit')` and `self.stat_count('fallback.cache.miss')`

hannosch:
I think we should extend this be one more thing: caching of not found responses from the fallback provider. This should still be limited to single-cell-only look-ups. Based on the early stats I'm seeing, we are getting repeat requests for the same cell-id in consecutive seconds or minutes.

We can store some kind of special token value in redis to signify a not found response.

jaredkerim:
@hannosch r?

ichnaea/locate/provider.py
```diff
@@ -609,24 +645,37 @@ def locate(self, data):
                     verify=False,
                 )
             self.stat_count('fallback.lookup_status.%s' % response.status_code)
-            if response.status_code == 404:
+            if response.status_code != 404:
                 # don't log exceptions for normal not found responses
-                response = None
-            else:
```
hannosch: @jaredkerim if you'd add this `response = None` back, that should result in the "cache not found" test to break, right?

What happens now is that we store the large nested JSON dict response from the fallback service into the cache. And than later on we try to parse that and get `response['location']['lat']` and friends from it, which fails and triggers a KeyError. That's not really all that nice and we trigger a raven error message each time. If you add a `self.check_raven(total=0)` to the test, that should fail right now and log this KeyError.

I'd rather make this explicit and create a special token value for not found and check for that one.

hannosch:
Looks all good, except for my comment at https://github.com/mozilla/ichnaea/pull/436/files#r30816625 - if you could address that, you can rebase/squash the entire thing and I'll merge it tomorrow.

jaredkerim:
A 404 from the fallback provider is cached as a boolean.  Tests updated with your suggestion to check for no raven errors.  Commits are squashed and rebased, should be good to merge.  r? @hannosch

Issue #438, opened 2015-05-22
Add explicit fallback choices to geolocate API


hannosch:
This is an extended spin-off from #421, where I said:

* In the fallback query, use the custom syntax and add `"fallbacks": {"lacf": 1}` to the geolocate query (see http://combain.com/api/ fallbacks or the WCDMA lacf example.

Instead of just explicitly doing this for the queries we are sending to the fallback provider, we should just add proper support for specifying fallbacks to our own query API and than pass through those fallback choices to the fallback provider.

This would allow someone to call our service and disable either the IP fallback and/or the cell area fallback.

This means our own geolocate query should get one more top-level mapping called `fallbacks`, on the same level as cellTowers/wifiAccessPoints. An example would be:

```
{
	"cellTowers": [{
		"radioType": "gsm",
		"mobileCountryCode": 240,
		"mobileNetworkCode": 1,
		"locationAreaCode": 3012,
		"cellId": 0
	}],
	"fallbacks": {
		"lacf": 1
	}
}
```
Combain does support a lot of different fallback options, as documented at http://combain.com/api/#1423734237921-5-5 (under the fallback parameter tab). Not all of those make sense for us, so I'd limit our support to just two:
* lacf (triggers using our cell_area/ocid_cell_area tables, and forwards the option to the fallback provider)
* ipf (use the IP address of the remote, don't support explicitly passing in a different IP address at this point), if disabled don't use our geoip provider to determine a location (might still be used to determine a country for metrics purposes).

In order to not change our current behavior, I'd default both options to true. As another complication, we don't want to send the ipf key in the outbound query to the fallback provider, but just the lacf key.

The second part of this is adding a `fallback` key to our response, for example:
```
{
	"location": {
		"lat": 59.393072,
		"lng": 18.160667
	},
	"accuracy": 46617,
	"fallback": "lacf"
}
```
This fallback key would only be set if the returned result is based on one of the fallbacks and if so is either set to lacf or ipf. The most complicated case here is that we'd end up using the location from the fallback provider and it was based on a lacf fallback, so our response should also say fallback:lacf.

I wouldn't add any of this to our old search API. In the request I'd accept both `0/1` and `true/false` as valid arguments.

jaredkerim:
Pre squashed r? @hannosch

PR #439, opened 2015-05-26
Add explicit fallback choices to geolocate API #438

jaredkerim:
- Accept lacf/ipf flags in search api
- Accept lacf/ipf flags in search providers
- Pass lacf/ipf flags to fallback provider
- Add tests

### DIFF
--- a/ichnaea/locate/provider.py
+++ b/ichnaea/locate/provider.py
@@ -60,6 +60,7 @@ class Provider(StatsLogger):
         The name to use in logging statements, for example 'cell_lac'
     """
 
+    fallback_field = None
     log_name = None
     location_type = None
     source = DataSource.Internal
@@ -70,7 +71,11 @@ class Provider(StatsLogger):
         self.geoip_db = geoip_db
         self.settings = settings
         self.redis_client = redis_client
-        self.location_type = partial(self.location_type, source=self.source)
+        self.location_type = partial(
+            self.location_type,
+            source=self.source,
+            fallback=self.fallback_field,
+        )
         super(Provider, self).__init__(*args, **kwargs)
 
     def should_locate(self, data, location):
@@ -79,6 +84,14 @@ class Provider(StatsLogger):
         found by another provider, check if this provider should
         attempt to perform a location search.
         """
+        if self.fallback_field is not None:
+            try:
+                fallbacks = data.get('fallbacks', {})
+                if fallbacks:
+                    return fallbacks.get(self.fallback_field, True)
+            except ValueError:
+                self.raven_client.captureException()
+
         return True
 
     def locate(self, data):  # pragma: no cover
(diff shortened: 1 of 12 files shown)


ichnaea/locate/provider.py
```diff
@@ -600,11 +610,11 @@ def _prepare_data(self, data):
...
-        query = {}
-        if cell_queries:
-            query['cellTowers'] = cell_queries
-        if wifi_queries:
-            query['wifiAccessPoints'] = wifi_queries
+        query = {
+            'cellTowers': cell_queries,
+            'wifiAccessPoints': wifi_queries,
+            'fallbacks': data.get('fallbacks', {}),
+        }
```
hannosch: I still prefer to not send empty arrays. If an inbound query didn't contain any cell data, we can just avoid sending `"cellTowers": []` in the outbound query, same for wifi or fallbacks.

ichnaea/service/search/schema.py
```diff
@@ -38,9 +38,16 @@ class WifisSchema(SequenceSchema):
...
+    ipf = SchemaNode(Integer(), missing=None)
+
+
 class SearchSchema(MappingSchema):
 
     radio = SchemaNode(String(),
                        validator=OneOf(RADIO_STRINGS), missing=None)
     cell = CellsSchema(missing=())
     wifi = WifisSchema(missing=())
+    fallbacks = FallbackSchema(missing={})
```
hannosch: A class level missing dict here is dangerous. I don't want to think about what this actually means in terms of sharing mutable global state. I'd rather make this None and deal with the dict transform later in the code.

ichnaea/service/search/schema.py
```diff
@@ -38,9 +38,16 @@ class WifisSchema(SequenceSchema):
     wifi = WifiSchema()
 
 
+class FallbackSchema(MappingSchema):
+
+    lacf = SchemaNode(Integer(), missing=None)
+    ipf = SchemaNode(Integer(), missing=None)
```
hannosch: Would it work to make this a boolean node? Does that convert `0/1` and JSON `true/false` to True/False?

hannosch:
This misses the schema changes to ichnaea.service.geolocate.schema (used by v1/geolocate and v1/country). It also send on all fallback choices to the fallback provider. The fallback provider should filter out the ipf field from the fallbacks, we never want to send that one to the fallback provider.

It also misses the second part, which is to return the used fallback in our responses (either because we used one of the cell area providers or because the fallback provider told us that it used some fallback).

ichnaea/service/search/schema.py
hannosch: And sometimes I wish Python had an immutable dict type, so you could actually specify a method level or class level dict-like default argument. Using a tuple as a placeholder for list works, but isn't ideal either.

hannosch:
r+, please squash, and feel free to merge.

One more thing: I tested the combain API, and they do accept JSON true/false values instead of 0/1. Personally I'd rather use true/false. If you want to, could you change the values over to booleans? Should be almost test-only changes, except for the fallback provider line where it says "We only send the lacf fallback for now".

Issue #440, opened 2015-05-28
Allow v1/v2 API queries without CID


jaredkerim:
Presently we have tests to ensure that Cell, Wifi, GeoIP, and Fallback are used in the v1 and v2 APIs, but we don't specifically test that CellArea's are able to satisfy requests for these APIs.  Add tests that specifically test against CellArea.

These should also encompass the case of satisfying successful geolocation queries with only radio, mcc, mnc, lac fields present (assuming they match a cell area in the database).

PR #441, opened 2015-05-28
Allow v1/v2 API queries without CID #440

jaredkerim:
- Create CellAreaLookup and ValidCellAreaKeySchema
- Abstract out schema validator from Cell providers
- Add tests to v1 and v2 APIs


jaredkerim:
@hannosch r?

Issue #443, opened 2015-06-02
Rename search, geolocate, submit, geosubmit, geosubmit2


jaredkerim:
We should rename our submit and locate modules to:

Search/search -> LocateV1/locate_v1
Geolocate/geolocate -> LocateV2/locate_v2
Submit/submit -> SubmitV1/submit_v1
Geosubmit/geosubmit -> SubmitV2/submit_v2
Geosubmit2/geosubmit2 -> SubmitV3/submit_v3

PR #445, opened 2015-06-02
Don't use external API names in internal class/module names.

hannosch:
This addresses #443 - I chose `locate` so it's not the same as the external API name, neither for `v1/geolocate` nor `v1/search`.

@jaredkerim r?


jaredkerim:
@hannosch I've already implemented this here:

https://github.com/JaredKerim-Mozilla/ichnaea/commit/4e3b081b26f79ca083094b0e82348b535483043b

Sorry, I should've assigned myself to the ticket so that it was clear I was already working on it.

PR #447, opened 2015-06-03
Rename search, geolocate, submit, geosubmit, geosubmit2 #443

jaredkerim:
Moved all the folders around and renamed classes.  Pre squash.

r? @hannosch


hannosch:
Overall looks ok, with the exception that I'd revert all the commits that change the full import path / dotted name of any of the task functions.

Mentally you have to think about these tasks in the same way you think of public HTTP API's or if this would be a reusable library, as public Python API's. You have to deprecate them, leave the old ones in place for one major release, add the new ones, wait for everyone to use the new API's and only afterwards remove the old ones. In our case "one major release" just happens to be one production push cycle, so it's often a lot shorter than a couple months/years. But these tasks still fall in the same category.

hannosch:
As for the monitor stuff, I very much like the service to api rename. I'd use the api sub-package for anything that touches actually public HTTP API's, those we document for end-users. And move all the internal bits (`__heartbeat__`/`__monitor__`) to the monitor sub-package.

Issue #444, opened 2015-06-02
Replace LocateTransform with an internal class based representation


jaredkerim:
Presently, we receive locate queries into both v1 and v2 in different data encoding schemas, and these are reconciled using the LocateTransform(ReportTransform) class which parses them into a consistently named set of dictionaries.  We would like to replace this with a set of schema specific deserializers which deserialize into an internal object representation (similar to a DB Model representation, but purely in memory).  Potentially overloading Colander, or just using NamedTuples, or something.

hannosch:
As of 377d419ea440b6b167c08b21793678721b0e0c7d this is actually done.

PR #450, opened 2015-06-28
Replace LocateTransform with an internal class based representation #444

jaredkerim:
This is still split into separate commits in case we need to isolate and revert any specific change, should make it easier to understand each step.  This adds the Internal Mapping types, converts all of the request schemas to use them, and removes the Transform objects.

@hannosch r?

### DIFF
--- a/ichnaea/api/locate/views.py
+++ b/ichnaea/api/locate/views.py
@@ -2,12 +2,9 @@ from pyramid.httpexceptions import HTTPNotFound
 from simplejson import dumps
 
 from ichnaea.api.locate.searcher import PositionSearcher
-from ichnaea.models.transform import ReportTransform
 from ichnaea.api.error import (
     JSONParseError,
-    preprocess_request,
 )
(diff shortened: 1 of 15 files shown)


jaredkerim:
Okay I think I've gotten all the feedback in and cleaned up the code style warnings.  I think it should be good to rebase, squash, and merge.  @hannosch r?

ichnaea/api/locate/views.py
```diff
@@ -94,3 +53,13 @@ def locate(self, api_key):
             api_key=api_key,
             api_name=self.view_name,
         ).search(query)
+
+    def prepare_location_data(self, location_data):
+        return location_data
```
hannosch: Can you stick a pragma no cover on this line, to get line coverage back to 100%? Not quite sure if this should be a raise NotImplementedError instead.

hannosch:
One nitpick (pragma) and one test problem left. The accept empty test stuff is ok as-is, follow-up bug for that is #451.

PR #522, opened 2018-06-22
fix line separator issues in public csv exporter

rindeal:
The header row was using '\n', while the spec and the rest of the code used '\r\n'.

### DIFF
--- a/ichnaea/data/public.py
+++ b/ichnaea/data/public.py
@@ -15,6 +15,8 @@ from ichnaea import util
 
 def write_stations_to_csv(session, path, today,
                           start_time=None, end_time=None):
+    linesep = '\r\n'
+
     where = 'lat IS NOT NULL AND lon IS NOT NULL'
     if start_time is not None and end_time is not None:
         where = where + ' AND modified >= "%s" AND modified < "%s"'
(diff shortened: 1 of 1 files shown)


rindeal:
bump

ichnaea/data/public.py
```diff
@@ -76,9 +78,7 @@ def write_stations_to_csv(session, path, today,
                             cellid=min_cellid
                         )).fetchall()
                     if rows:
-                        buf = '\r\n'.join([row.cell_value for row in rows])
-                        if buf:
-                            buf += '\r\n'
+                        buf = linesep.join([row.cell_value for row in rows]) + linesep
```
jwhitlock: This line is now too long (86 characters).

ichnaea/data/public.py
```diff
@@ -15,6 +15,15 @@
 
 def write_stations_to_csv(session, path, today,
                           start_time=None, end_time=None):
+    linesep = '\r\n'
+
+    header_row = [
```
jwhitlock: This section should not have been moved, it makes it harder to understand what changed.

rindeal:
Thanks for your review, I have incorporated the changes you requested. Everything should be well now.

hannosch:
> I would have switched to using the [csv module](https://docs.python.org/2/library/csv.html) instead of manually creating CSV, and still might after this lands.

This used the csv module at some point. But performance was killed by having the Python database driver do binary network packet to Python type conversions on those 46 million rows multiplied by 12 columns. This jobs results in 1.5gb of uncompressed/500mb of gzip compressed data, when it is run once per day. Streaming the result in chunks and avoiding many tiny packets helped to keep total runtime and memory overhead in check. The code is far from pretty as a result.

ichnaea/data/public.py
```diff
@@ -76,9 +78,7 @@ def write_stations_to_csv(session, path, today,
                             cellid=min_cellid
                         )).fetchall()
                     if rows:
-                        buf = '\r\n'.join([row.cell_value for row in rows])
-                        if buf:
-                            buf += '\r\n'
+                        buf = ''.join(row.cell_value + linesep for row in rows)
```
jwhitlock: It took me a minute to figure this out (I had to go back and understand ``row.cell_value``), but it looks like it will do the same thing as the previous code.

jwhitlock:
Thanks @hannosch for that info on using ``csv`` for this. That may save us some wasted effort down the road.

PR #796, opened 2019-06-04
[do not merge] finish circle-ci migration

lonnen:
the circleci migration tool for pre-2.0 ~> 2.* formats examines the disk to determine if everything is cool, instead of looking at the repo contents or requesting the contents of the remote. Because of a blanket dotfile line in the .gitignore the necessary config was never sent up to the server and builds continue failing

this fixes that line and adds the auto-generated config file into version control


lonnen:
obviated by #821 and #822

PR #814, opened 2019-06-20
Disable scheduled pyup updates

jwhitlock:
Instead of opening a pyup PR every day, open one for each dependency. This should reduce the traffic from a new PR every day, plus the noise from auto-closed PRs.


hannosch:
+1 from me

lonnen:
:wave: @hannosch hope you're well

thanks @jwhitlock

PR #819, opened 2019-07-09
Suspending API key issuance

ckolos:
add verbiage about suspension of API key issuance

### DIFF
--- a/ichnaea/content/templates/api.pt
+++ b/ichnaea/content/templates/api.pt
@@ -61,6 +61,12 @@
     </p>
 
     <p class="text-justified">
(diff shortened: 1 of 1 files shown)


ichnaea/content/templates/api.pt
```diff
@@ -60,6 +60,12 @@
         service.
```
jasonthomas: Should the paragraph above also be commented out?

ckolos: So I thought about that, I'm not sure it should be removed only b/c it describes how the api functions wrt the app itself.

PR #821, opened 2019-07-29
Circle CI 2.0 migration, attempt #2

jwhitlock:
Build on PR #796, to get the 2.0 configuration working.


jwhitlock:
Trying to close / open PR to test new CircleCI setting (build PRs from forks without secrets)

jwhitlock:
This is ready for review.

One of the changes to get tests running is to add SSL connection parameters when using [mysql-connector-python](https://pypi.org/project/mysql-connector-python/). The default in Debian 10 (buster) appears to be TLS v1.2, but this isn't supported by our MySQL deployments:
* [AWS MySQL 5.6](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_MySQL.html#MySQL.Concepts.SSLSupport) supports TLS v1.0, but not v1.1 or v1.2
* The MySQL 5.7 Docker image supports TLS v1.0 and v1.1, but [not v1.2](https://github.com/docker-library/mysql/issues/567).

So, this sets the protocol version to TLS v1.0. Alternatively, we can explicitly turn SSL off for MySQL Connector, which may have been the default in 2.2.3. 

A path to a certificate authority file was also needed, so I used the one from [certifi](https://pypi.org/project/certifi/).

I think there is still work to do to bring over the deploy actions from the v1 config, but at least we have a working build, so progress.

.circleci/config.yml
```diff
@@ -0,0 +1,74 @@
...
+                I="image-$(date +%j).gz"
+                if [[ -e ~/docker/$I ]]
+                    then echo "Loading $I"
+                    pigz -d -c ~/docker/$I | docker load
+                fi
+                docker build -t ${DOCKERHUB_REPO:-mozilla/location} .
+                mkdir -p ~/docker
+                rm -f ~/docker/*
+                docker save ${DOCKERHUB_REPO:-mozilla/location} | pigz --fast -c > ~/docker/$I
+                ls -l ~/docker
```
willkg: This is a bunch of interesting code to cache intermediary Docker layers to speed up Circle CI runs. One thing that concerns me is that it presumes that we have `Dockerfile` set up in such a way that it correctly breaks the cache when appropriate. I find that to be a common source of errors and it's tricky to test and verify it's correct.

I would simplify this and nix the layer-caching code. We can re-add it if build times are significant.

hannosch: +1 this caching code is ancient. Simplify first, speedup later.

jwhitlock:
I agree, we should simplify the configuration down to the essentials before building it up again. My experience is that circleci config changes take a few iterations (even with ``circleci local execute``), so I'm taking my two +r and will work on the next steps in future PRs:

* Remove the docker image caching
* Simplify other job steps
* Remove non-working test results upload
* Add steps to upload images to docker
* Ensure omitted branches remain omitted

jwhitlock:
We've also confirmed that the web app uses the PyMySQL library in stage and prod, so this extra SSL code for MySQL Connector/Python should not run. More research is needed. [Django](https://docs.djangoproject.com/en/2.2/ref/databases/#mysql-db-api-drivers) recommends [mysqlclient](https://pypi.org/project/mysqlclient/), followed by MySQL Connector/Python. We should see if our libraries have recommendations, as we get to library upgrades.

hannosch:
> We've also confirmed that the web app uses the PyMySQL library in stage and prod, so this extra SSL code for MySQL Connector/Python should not run. More research is needed. [Django](https://docs.djangoproject.com/en/2.2/ref/databases/#mysql-db-api-drivers) recommends [mysqlclient](https://pypi.org/project/mysqlclient/), followed by MySQL Connector/Python. We should see if our libraries have recommendations, as we get to library upgrades.

The webapp uses gevent + pymysql as back when this was written, the gevent monkey patches didn't work against C libraries.

The celery worker backend and the command line scripts / cron jobs use a C-based DB library to improve performance. Those use multiprocessing with separate processes and no gevent.

The particular DB libraries were based off whatever SQLAlchemy recommended at the time.

willkg:
@hannosch That's really helpful to know--thank you!

PR #822, opened 2019-07-30
Not Ready: Circle CI update, round 3

jwhitlock:
Building on PR #821:

* [ ] Collect test results
* [ ] Ignore the same branches
* [ ] Upload images to dockerhub for select branches


jwhitlock:
I've hit a dead end on generating and uploading the jUnit-style XML needed for CircleCI to report detailed test results. The file is generated inside the container, but I'm unable to mount a volume such that the ``./dev test`` command will place the file where CircleCI can retrieve it. 

Instead of fixing broken CI features, I'm going to drop them, get to a simple config, and then add the features back if needed. Some of that simplification should be done on PR #821.

PR #825, opened 2019-08-02
Document requirements, split constraints

jwhitlock:
Determine the primary requirements, using [pipdeptree](https://github.com/naiquevin/pipdeptree) and looking for ``import`` statements in the code. For each primary requirement, add a one-line summary for each, and link to the source repo. Move the requirements of the primary requirements to a new ``constraints.txt`` file.

Reformat and complete the hash list for each with [hashin](https://github.com/peterbe/hashin).

This should just be a reformatting of the files, with no requirement updates or removals. The next step would be to update some of them. I did change some package names and version numbers, to use the capitalization and punctuation expected by `hashin`. I'll add a self-review to point these out, as well as some future work.


requirements/all.txt
```diff
@@ -110,14 +64,37 @@ Cython==0.26 \
...
-geoip2==2.5.0 --hash=sha256:198fc36b51dd757df02173862a7ba48c2f61ee416c26e782cd8a887fd9aedb41
+
+# SDK for metrics aggregation server
+# Code: https://github.com/DataDog/datadogpy
+datadog==0.16.0 \
+    --hash=sha256:9ce3c3c4080f6dcb9c799d3d9e43ea3ed61132599d0e9143954d35ae9578a2d1
+
+# Test fixtures replacement
+# Code: https://github.com/FactoryBoy/factory_boy/
+factory-boy==2.9.2 \
```
jwhitlock: was ``factory_boy`` with an underscore

requirements/all.txt
```diff
@@ -148,187 +125,69 @@ gevent==1.2.2 \
...
+# "Green Unicorn", a WSGI HTTP Server
+# Code: https://github.com/benoitc/gunicorn
+gunicorn==19.7.1 \
+    --hash=sha256:75af03c99389535f218cc596c7de74df4763803f7b63eb09d77e92b3956b36c6 \
+    --hash=sha256:eee1169f0ca667be05db3351a0960765620dad53f53434262ff8901b68a1b622
+
+# Python wrapper for a minimalistic C client for Redis
+# TODO: redis or hiredis or both?
+# Code: https://github.com/redis/hiredis-py
+hiredis==0.3.1 \
```
jwhitlock: ``hiredis`` appears to be unused

requirements/all.txt
```diff
@@ -148,187 +125,69 @@ gevent==1.2.2 \
...
+    --hash=sha256:d8222d09da762891d4ef39374b1f2318da0ba02faf609d254438672892155eee \
+    --hash=sha256:dac079ff66c2d1c56f62f0a9b0a7a01784ad7b9eb136b9a931f7ab57463e70de \
+    --hash=sha256:e68638e460e95a68dbcef6f78be8210f4405eac020f10ee52cafee54084d3699 \
+    --hash=sha256:fb8b80d2fb45857fd19d04387e4207412124e470c3b5d4d3568a6d9b74a72e99 \
+    --hash=sha256:fc3eb53b51acf467ae7d60e92af23dee33b20871d84ce33e6e61b9e00077ff48 \
+    --hash=sha256:ffe6d9bc7f3637d962eb4706a2908ba4064ec997beb99540bb385a0f7d7fe39a
+
+# Messaging library
+# Code: https://github.com/celery/kombu
+kombu==3.0.37 \
```
jwhitlock: I was surprised that ``kombu`` is used directly in the code. On other projects, Celery requires it but not the app code.

requirements/all.txt
```diff
@@ -148,187 +125,69 @@ gevent==1.2.2 \
...
+    --hash=sha256:e064a00c66b4d1058cd2b0523fb8d98c82c18450244177b6c0f7913016642650
+
+# Python MaxMind DB reader extension
+# Code: https://github.com/maxmind/MaxMind-DB-Reader-python
+maxminddb==1.3.0 \
+    --hash=sha256:972399b6d08b4cbb7889f2d2489266d18811c5ef6109cc52eb28de4102bbc4be
+
+# Mobile Country Codes (unmaintained)
+# Code: https://github.com/andymckay/mobile-codes
+mobile-codes==0.7 \
```
jwhitlock: [andymckay/mobile-codes](https://github.com/andymckay/mobile-codes) is marked as unmaintained.

requirements/all.txt
```diff
@@ -358,3 +217,208 @@ numpy==1.15.1 \
     --hash=sha256:df0b02c6705c5d1c25cc35c7b5d6b6f9b3b30833f9d178843397ae55ecc2eebb \
     --hash=sha256:e3660744cda0d94b90141cdd0db9308b958a372cfeee8d7188fdf5ad9108ea82 \
     --hash=sha256:f2362d0ca3e16c37782c1054d7972b8ad2729169567e3f0f4e5dd3cdf85f188e
+
+# Python Build Reasonableness for setuptools
+# TODO: Try removing
+# Code: https://github.com/openstack/pbr
+pbr==3.1.1 \
```
jwhitlock: ``pbr`` doesn't appear to be used directly. It may patch setuptools in ways we need.

requirements/all.txt
```diff
@@ -358,3 +217,208 @@ numpy==1.15.1 \
...
+    --hash=sha256:cc5eb13f5ccc4b1b642cc147c2cdd121a34278b341c7a4d79e91182fff425836 \
+    --hash=sha256:cd3b0e0ad69b74ee55e7c321f52a98effed2b4f4cc9a10f3683d869de00590d5 \
+    --hash=sha256:d6e88c4920660aa75c0c2c4b53407aef5efd9a6e0ca7d2fc84d79aba2ccbda3a \
+    --hash=sha256:ec3c49b6d247152e19110c3a53d9bb4cf917747882017f70796460728b02722e \
+    --hash=sha256:f1f5d8b8e0bc9651d81b40ad3d9fb7cdd858ea31fc116dd230393465849dbecd
+
+# Generic ASN.1 library
+# TODO: Try removing
+# Code: https://github.com/etingof/pyasn1
+pyasn1==0.3.2 \
```
jwhitlock: ``pyasn1`` doesn't appear to be used by our code.

requirements/all.txt
```diff
@@ -358,3 +217,208 @@ numpy==1.15.1 \
...
+
+# A Python web framework
+# Code: https://github.com/Pylons/pyramid
+pyramid==1.9.1 \
+    --hash=sha256:7da99f42fbc31cb36ae5297091734cab4e27d222da806dd370f3466ea8c6c422 \
+    --hash=sha256:840f4bbd5994609340de35079da0d1dd94f79942e27818d5d8566f8009fb0b36
+
+# Pyramid bindings for Chameleon, an HTML/XML template engine
+# Code: https://github.com/Pylons/pyramid_chameleon
+pyramid_chameleon==0.3 \
```
jwhitlock: Was ``pyramid-chameleon``, with a dash

requirements/all.txt
```diff
@@ -358,3 +217,208 @@ numpy==1.15.1 \
...
+    --hash=sha256:78713101b32af384c564837fd7ae665a2a72cb6d49edbd8f32148d74724a65a8 \
+    --hash=sha256:909be30bdd1d8cd57609760ea72a1499543169ed6ea5c66fad50946582682cf8 \
+    --hash=sha256:93825cc80c638d901099f657dfff852ad2421beb51cb7d1d3f91157741ebe287 \
+    --hash=sha256:a19a2ca7a7336495ec180adeaa0dfdcf41e96dbbee90d51c3ed828ba570884e6 \
+    --hash=sha256:b3e97be2cd9f052d984fc5ba2d441897971b744c64d658617944c47bc366f8ff \
+    --hash=sha256:b87a923ed390ba0aafcbbcc6f0e023d495f19d2bd22ae59bdef3b0aec6485b39
+
+# Customize the process title
+# https://github.com/dvarrazzo/py-setproctitle
+setproctitle==1.1.10 \
```
jwhitlock: ``setproctitle`` may be unused

requirements/all.txt
```diff
@@ -358,3 +217,208 @@ numpy==1.15.1 \
...
+
+# Documentation builder
+# Code: https://github.com/sphinx-doc/sphinx/
+Sphinx==1.6.3 \
+    --hash=sha256:3ea0faf3e152a0e40372d8495c8cbd59e93f89266231c367d8098ec0dfede98f \
+    --hash=sha256:af8bdb8c714552b77d01d4536e3d6d2879d6cb9d25423d29163d5788e27046e6
+
+# ReadTheDocs theme for Sphinx
+# Code: https://github.com/readthedocs/sphinx_rtd_theme
+sphinx-rtd-theme==0.2.4 \
```
jwhitlock: Was ``sphinx_rtd_theme``, changed to dashes

requirements/all.txt
```diff
@@ -1,86 +1,40 @@
...
+
+# Curated collection of root certificates
+# Code: https://github.com/certifi/python-certifi
+certifi==2017.7.27.1 \
+    --hash=sha256:40523d2efb60523e113b44602298f0960e900388cf3bb6043f645cf57ea9e3f5 \
+    --hash=sha256:54a07c09c586b0e4c619f02a5e94e36619da8e2b053e20f594348c0611803704
+
+# Schema-based serialization and deserialization library
+# Code: https://github.com/Pylons/colander
+colander==1.4 \
```
jwhitlock: Was version ``1.4.0``, but just ``1.4`` on PyPI

requirements/constraints.txt
```diff
@@ -0,0 +1,351 @@
...
+# alembic, Sphinx
+MarkupSafe==1.0 \
+    --hash=sha256:a6be69091dac236ea9c6bc7d012beab42010fa914c459791d627dad4910eb665
+
+# alembic, boto3, factory_boy, ndg_httpsclient
+python-dateutil==2.6.1 \
+    --hash=sha256:891c38b2a02f5bb1be3e4793866c8df49c7d19baabf9c1bad62547e0b4866aca \
+    --hash=sha256:95511bae634d69bc7329ba55e646499a842bc4ec342ad54a8cdb65645a0aad3c
+
+# alembic, boto3, factory_boy, ndg_httpsclient, protobuf, Sphinx
```
willkg: I don't know if I'd keep these comments. They're nice now because you did a lot of research, but these things change frequently and there's no good way to know when it changed and update the comment accordingly.

requirements/all.txt
```diff
@@ -1,86 +1,40 @@
...
-    --hash=sha256:8168bcb08403ef144ff1fb880d416f49e2728101d02aaadfe9645883222c0aa5 \
-    --hash=sha256:8229ceb79a1792823d87779959184a1bf95768e9248c93ae9f97c7a2f60376a1 \
-    --hash=sha256:8a19e9f2fe69f6a44a5c156968d9fc8df56d09798d0c6a34ccc373bb186cee86 \
-    --hash=sha256:8d10113ca826a4c29d5b85b2c4e045ffa8bad74fb525ee0eceb1d38d4c70dfd6 \
-    --hash=sha256:be495b8ec5a939a7605274b6e59fbc35e76f5ad814ae010eb679529671c9e119 \
-    --hash=sha256:dc2d3f3b1548f4d11786616cf0f4415e25b0fbecb8a1d2cd8c07568f13fdde38 \
-    --hash=sha256:e4aecdd9d5a3d06c337894c9a6e2961898d3f64fe54ca920a72234a3de0f9cb3 \
-    --hash=sha256:e79ab4485b99eacb2166f3212218dd858258f374855e1568f728462b0e6ee0d9 \
-    --hash=sha256:f995d3667301e1754c57b04e0bae6f0fa9d710697a9f8d6712e8cca02550910f
+-c constraints.txt
```
willkg: We should do a FIXME to rename this file. It's not really "all the requirements" anymore.

requirements/binary.txt
```diff
@@ -1 +1,22 @@
-mysql-connector-python==8.0.16 --hash=sha256:ea7d80e2b406dba3b20836e95d76e926d0f80034e5a015c0d8309d820225cb5b
\ No newline at end of file
+# MySQL driver written in Python
+# Code: https://github.com/mysql/mysql-connector-python
+mysql-connector-python==8.0.16 \
```
jwhitlock: I'm not sure why this isn't in ``all.txt``. It is possible that 8.x no longer requires the MySQL C libraries.

requirements/constraints.txt
```diff
@@ -0,0 +1,351 @@
...
+hupper==1.0 \
+    --hash=sha256:afb9584fc387c962824627bb243d7d92c276df608a771b17c8b727a7de34920a \
+    --hash=sha256:e0dbe89b637b59d82687ba7f74690a0fa37d0f5ca4963cbb9069834ab08272fe
+PasteDeploy==1.5.2 \
+    --hash=sha256:39973e73f391335fac8bc8a8a95f7d34a9f42e2775600ce2dc518d93b37ef943 \
+    --hash=sha256:d5858f89a255e6294e63ed46b73613c56e3b9a2d82a42f1df4d06c8421a9e3cb
+plaster==0.5 \
+    --hash=sha256:2a028938dcbf41033c5d377363781b2528151b0159201587c41e7a4c74bc887c \
+    --hash=sha256:626f4c2ffaaa101b9eced20da42fe44c460f1b0bb52b0add37b4836516978f53
+plaster-pastedeploy==0.4.1 \
```
jwhitlock: Was ``plaster_pastedeploy``, changed to a dash

requirements/constraints.txt
```diff
@@ -0,0 +1,351 @@
...
+Jinja2==2.9.6 \
+    --hash=sha256:2231bace0dfd8d2bf1e5d7e41239c06c9e0ded46e70cc1094a0aa64b0afeb054 \
+    --hash=sha256:ddaa01a212cd6d641401cb01b605f4a4d9f37bfc93043d7f760ec70fb99ff9ff
+Pygments==2.2.0 \
+    --hash=sha256:78f3f434bcc5d6ee09020f92ba487f95ba50f1e3ef83ae96b9d5ffa1bab25c5d \
+    --hash=sha256:dbae1046def0efb574852fab9e90209b23f556367b5a320c0bcb871c77c3e8cc
+snowballstemmer==1.2.1 \
+    --hash=sha256:919f26a68b2c17a7634da993d91339e288964f93c274f1343e3bbbe2096e1128 \
+    --hash=sha256:9f3bcd3c401c3e862ec0ebe6d2c069ebc012ce142cce209c098ccb5b09136e89
+sphinxcontrib-websupport==1.0.1 \
```
jwhitlock: was ``sphinxcontrib_websupport``, changed to a dash

requirements/all.txt
```diff
@@ -1,86 +1,40 @@
...
+    --hash=sha256:6ced63033bc663e60c992564954dbb5c84c43899f7f1a04b739957350f6b55f3
+
+# Curated collection of root certificates
+# Code: https://github.com/certifi/python-certifi
+certifi==2017.7.27.1 \
+    --hash=sha256:40523d2efb60523e113b44602298f0960e900388cf3bb6043f645cf57ea9e3f5 \
+    --hash=sha256:54a07c09c586b0e4c619f02a5e94e36619da8e2b053e20f594348c0611803704
+
+# Schema-based serialization and deserialization library
+# Code: https://github.com/Pylons/colander
```
willkg: I'm not sure I'd keep these comments, either. Most of these are easy lookups in PyPI. The data here won't change as much as `constraints.txt`, but I don't think I'd want to maintain them.

requirements/build.txt
```diff
@@ -1,6 +1,72 @@
-pip==9.0.1 --hash=sha256:690b762c0a8460c303c089d5d0be034fb15a5ea2b75bdf565f40421f542fefb0
-setuptools==36.2.7 --hash=sha256:02b5f3869777dfbfbe3786a057bc6d523445a3e541a0abdf85adbe0b091e96b9
-wheel==0.29.0 --hash=sha256:ea8033fc9905804e652f75474d33410a07404c1a78dd3c949a66863bd1050ebd
-numpy==1.13.1 --hash=sha256:d910a24f536f926bd56fb30d6f17ae8b89a1406e105087a49e014e000b00e8db
-pycparser==2.18 --hash=sha256:99a8ca03e29851d96616ad0404b4aad7d9ee16f25c9f9708a11faf2810f7b226
-cffi==1.10.0 --hash=sha256:e7175287f7fe7b1cc203bb958b17db40abd732690c1e18e700f10e0843a58598
+pip==9.0.1 \
+    --hash=sha256:09f243e1a7b461f654c26a725fa373211bb7ff17a9300058b205c61658ca940d \
+    --hash=sha256:690b762c0a8460c303c089d5d0be034fb15a5ea2b75bdf565f40421f542fefb0
```
willkg: Don't we pick up `pip` from the Docker image? Do we need to install it, too?

willkg:
Out of curiosity, what's the set of commands you're iterating over to test this out?

jwhitlock:
I'm installing in a Python 3.6 virtualenv on my host, running pipdeptree and hashin there, etc. I'm using ``ack --python import | grep -v "from ichnaea"`` to get a long list of imports to comb through.

I used ``./dev test`` to rebuild the images and confirm things are still working.

requirements/build.txt
jwhitlock: Yep, I think this is a candidate for removal as well. There was a big jump in ``pip`` versions, they are up to 19.2.1.

requirements/constraints.txt
jwhitlock: OK, I'll remove them. Should ``constraints.txt`` be grouped by package, sorted by package name, or something else?

requirements/all.txt
jwhitlock: yeah, warning bells should go off when someone uses "all" or "new" or "old". I think it will require changing the Dockerfile or other build steps to change the name.

requirements/all.txt
jwhitlock: OK, I'll move to my notes and alpha-sort the packages

requirements/constraints.txt
willkg: My experience with Socorro leads me to believe that there's no good grouping that isn't hard to maintain, so I'd just let it sort itself however it likes. Hopefully there are better tools for managing things like this some day.

jwhitlock:
As requested, I've dropped the notes and URLs to the source repos.

requirements/all.txt
hannosch: This is an optional dependency of redis, implementing some of the code in C. Not sure if there was a setuptools extra back than, but today there is: https://github.com/andymccurdy/redis-py/blob/3afa016029d7e0a46fa1d76ac9c7463a02495e05/setup.py#L42

requirements/all.txt
hannosch: This is an optional dependency of gunicorn: https://github.com/benoitc/gunicorn/blob/0ce857fced9e8a66c05014054821ace9c56681b9/docs/source/faq.rst

requirements/binary.txt
hannosch: IIRC the split into different requirement files was based on doing things in different docker layers, so expensive an rarely changing libraries with C dependencies could be cached longer in local dev environments. But this was before binary wheels became a real thing. So probably this is a thing of the past, which was never updated.

requirements/all.txt
hannosch: IIRC this was a build dependency of something in the past, when you installed the library from a source dist. But I can’t remember anymore what that was.

requirements/all.txt
hannosch: This was a dependency of the cryptography library at the time.

requirements/all.txt
jwhitlock: Thanks, that explains what is going on. The ``redis`` devs added the ``extras_require`` in 2.10.6, so after updating we can switch to ``redis[hiredis]==2.10.6`` (or whatever version) and ``hiredis`` can move to constraints.

requirements/all.txt
jwhitlock: Thanks, that matches my vague memory of this module from past projects. I still see "Pabst Blue Ribbon", never "Python Build Reasonableness".

requirements/all.txt
hannosch: This is a bit of registration, test support and changing the serialization format to json using simplejson. At the time kombu still used unsafe pickles by default and the json support it had wasn’t using a particularly fast library. This project started out on Python 2.6 were the standard library json module had no C speedups.

requirements/all.txt
jwhitlock: Interesting, yes we want to keep that. It looks like gunicorn also has ``gunicorn[gevent]`` in the current release, but not our version.

requirements/all.txt
jwhitlock: That makes sense. I hadn't looked too closely at the usage yet, I was just focusing on keeping directly-used modules in ``requirements/all.txt``.

The Celery maintainers have caught up in the meantime. In the latest versions, JSON is the default, and you have to ignore the documented warnings to use pickle.

requirements/all.txt
jwhitlock: After a weekend, I'm thinking that I could include renaming this file in the PR. I could move one or two requirements per commit, which would let me use the commit comments to document my research and @hannosch's feedback on why we have requirements like ``setproctitle``.

The change to requirements names would affect about 3 other files. In ``docker.make``, it would change the command to build dependencies:

https://github.com/mozilla/ichnaea/blob/605776dea41aed748efe5a65ee6d8c8951444976/docker.make#L53-L58

It would change ``pyup.yml``, to change the names of the files to scan:

https://github.com/mozilla/ichnaea/blob/605776dea41aed748efe5a65ee6d8c8951444976/.pyup.yml#L3-L9

And finally, the docs would need some tweaking.

I think after migration, there will be two requirements files. I like ``constraints.txt`` for the constraints. Any preferences for the name of the other one? Here are some options:

* ``requirements/default.txt`` - Socorro, Antennta, other services projects
* ``requirements/primary.txt`` - To match how I talked about it in the PR

jwhitlock:
I'm replacing this with PR #826, which drops the unused requirements, and consolidates everything into ``default.txt`` and ``constraints.txt``.

Issue #831, opened 2019-08-07
get travis ci working with ckolos/dammit branch


willkg:
We're currently deploying MLS from the ckolos/dammit branch. Travis CI is set up for that branch, but failing:

https://travis-ci.org/mozilla/ichnaea/builds/557384813

This covers getting Travis CI working for that branch.

willkg:
@jwhitlock and I talked a bit about this. We can ditch Python 2.6 testing--that solves one failure. If we can get it passing with Python 2.7, that'd be great. If not, we should figure out another way to run the tests for changes made in this branch.

jwhitlock:
Production is running Python 2.6, so we should keep testing that if possible.

willkg:
Great googley moogely.

jwhitlock:
This is working with #833, and is now switched to looking at 1.5.x. We need to test deployment from 1.5.x branch, but that feels like a different issue.

PR #833, opened 2019-08-07
Get TravisCI building on ckolos/dammit branch

jwhitlock:
Get the 1.5.1 build in ckolos/dammit (the deployment branch) working again:

- Use trusty build
- Disable the built-in boto configuration
- Ensure MySQL starts
- Cherry-pick the MaxMindDB update from d3e8983a, minus the 2.x vendored stuff
- Add a badge to the README

This should fix issue #831.


hannosch:
Btw. There is a 1.5.x branch, which is almost the same as ckolos/dammit. The former was meant to be the old production branch, while master got dockerized and got some css changes to adopt the “new” Mozilla logo and style.

jwhitlock:
@ckolos @willkg how do you feel about using the [1.5.x branch](https://github.com/mozilla/ichnaea/tree/1.5.x) as the deployment branch rather than `ckolos/dammit`?

willkg:
I'm inclined to stick with `ckolos/dammit` unless you feel confident that you can do the switch with low risk. We're probably not going to touch this branch after September, so I don't feel compelled to clean it up. Beyond that, I don't have any strong opinions and I'm +1 on whatever you choose--you know more about the specifics than I do.

ckolos:
We'll need to merge over the verbiage changes from dammit (smh, if I had known we'd be using this I would have used a better name), but I'm more than willing to try a deploy from the 1.5.x branch to test in stage.

jwhitlock:
I'm going to merge this, then open a PR back to 1.5.x

PR #835, opened 2019-08-09
Merge 2018-2019 changes back to 1.5.x branch

jwhitlock:
The goal with this PR is to make ``1.5.x`` the canonical branch for deployment in the existing infrastructure. In the future, PRs destined for current deployment should use ``1.5.x`` as their base.


jwhitlock:
TravisCI is green, r=self

PR #841, opened 2019-08-14
local dev environment basics

### LINKED ISSUE #828: docker-based local development environment
With Socorro, Tecken, and Buildhub2, we were able to get a complete local development environment running with four or five commands (depending on what you were doing). This was based on make, git, docker, and docker-compose. It covered building images, initializing services, and fetching any initial data. It made it much easier to do development because we could throw away environments and start over trivially.

Properties I'd like in a local dev environment:

1. easy-to-document "getting started" for a local dev environment
2. easy to throw away all state and start over
3. easy to script environment building and testing

This issue covers getting a local dev environment set up for Ichnaea with similar properties.

willkg:
This reworks a bunch of things to create a local dev environment that's like the one we had with Tecken and Socorro. Further, it fixes some of the dockerization so it'll work better in GCP.

This is a work in progress and kind of a mess at the moment. I'll continue honing it over the next week.

Fixes #828.


willkg:
I need to fix tests and a bunch of stuff. I'll write up a list.

willkg:
This needs:

1. docs (me)
2. self-review (me)
3. regular review

I'll tag @jwhitlock after I've done the first two. There are still a bunch of things we could do, but this PR is getting long and we can push them off to future issues/PRs.

Dockerfile
```diff
@@ -1,31 +1,33 @@
-FROM python:3.6-slim
+FROM python:3.6.9-slim
```
willkg: I like pegging at a specific Python version so that it doesn't update underneath us.

Dockerfile
```diff
@@ -1,31 +1,33 @@
...
-# the application
-RUN groupadd -g 10001 app && \
-    useradd -d /app -g 10001 -G app -M -s /bin/sh -u 10001 app
+# Set up user and group.
+ARG groupid=10001
+ARG userid=10001
 
 WORKDIR /app
+RUN groupadd --gid $groupid app && \
+    useradd -g app --uid $userid --shell /usr/sbin/nologin --create-home app
```
willkg: This is how we did Socorro. This makes it easier for Linux developers to do development without dealing with file permission issues.

Dockerfile
```diff
@@ -1,31 +1,33 @@
...
 " > /etc/dpkg/dpkg.cfg.d/apt-no-docs
 
-# Install GNUPG
+# Install GNUPG.
 RUN apt-get update && apt-get -y install gnupg
 
-# Add MySQL apt repo & GPG key
-COPY ./conf/mysql_pubkey.asc /tmp/mysql_pubkey.asc
+# Add MySQL apt repo & GPG key.
+COPY ./docker/mysql_pubkey.asc /tmp/mysql_pubkey.asc
```
willkg: I moved all things related to building images into the `docker/` subdirectory including this pubkey.

Dockerfile
```diff
@@ -61,26 +63,21 @@ RUN make -f docker.make build_deps
 COPY ./requirements/*.txt /app/requirements/
 RUN make -f docker.make build_python_deps
 
-# Install the application code.
+# Install geocalc.
 COPY . /app
-RUN make -f docker.make build_ichnaea
+RUN make -f docker.make build_geocalc
```
willkg: Previously, this would build and install ichnaea which included building and installing the geocalc extension. I moved the geocalc code into a separate Python package, so that package gets built and installed, but the ichnaea package no longer does.

That fixes the issue where you had to install ichnaea in order to see changes you're making. It also fixes the issue where building geocalc creates build artifacts in `/app` which then get shadowed when the repository directory is mounted as `/app`.

Makefile
```diff
@@ -0,0 +1,127 @@
+# Include my.env and export it so variables set in there are available
```
willkg: Most of this file is boilerplate from Socorro.

Makefile
```diff
@@ -0,0 +1,127 @@
...
+	@echo "  mysql            - open mysql prompt"
+	@echo "  clean            - remove all build, test, coverage and Python artifacts"
+	@echo "  lint             - lint code"
+	@echo "  test             - run unit tests"
+	@echo "  testshell        - open a shell in the test environment"
+	@echo "  docs             - generate Sphinx HTML documentation, including API docs"
+	@echo "  buildjs          - generate js static assets"
+	@echo "  buildcss         - generate css static assets"
+	@echo ""
+	@echo "  setup            - set up mysql, redis, etc"
```
willkg: Oops--this is in here twice. I can nix this line.

Makefile
```diff
@@ -0,0 +1,127 @@
...
+
+.docker-build:
+	make build
+
+.PHONY: build
+build: my.env
+	${DC} build ${DOCKER_BUILD_OPTS} \
+	    --build-arg userid=${ICHNAEA_UID} \
+	    --build-arg groupid=${ICHNAEA_GID} \
+	    node app web scheduler
```
willkg: This should include worker.

willkg: Actually, on second thought, web, scheduler, and worker all use the same image, so we can nix those three.

bin/test_env.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+
```
willkg: This comes from Socorro. It gets around the issue where Circle CI can run docker-compose, but services can't mount volumes. We have to create a data container and copy the data into it and then mount that all without docker-compose.

Because I like running the tests in the local development environment in the same way that they're run in Circle CI, we do this mildly complicated song-and-dance. Further, I like having a test shell environment, so this enables that, too.

docker-compose.yml
```diff
@@ -1,15 +1,88 @@
...
+    env_file:
+      - docker/config/app.env
+      - docker/config/test.env
+      - my.env
+    depends_on:
+      - mysql
+      - redis
+    volumes:
+      - .:/app
+    command: ["shell"]
```
willkg: I redid how this works, so we don't need this anymore.

Makefile
```diff
@@ -0,0 +1,127 @@
...
+test: my.env .docker-build
+	./bin/test_env.sh
+
+.PHONY: testshell
+testshell: my.env .docker-build
+	./bin/test_env.sh --shell
+
+.PHONY: docs
+docs: my.env .docker-build
+	${DC} run --rm --no-deps test shell ./docker/run_build_docs.sh
```
willkg: This should use the `app` service--not the `test` service.

docker.make
```diff
@@ -1,10 +1,8 @@
 # This makefile is executed from inside the docker container.
```
willkg: We can probably nix this make file and turn it into bash scripts. That's how we do it in Socorro. This make file doesn't have any dependencies for its targets, so it's not getting anything out of being a make file.

docker/app_entrypoint.sh
```diff
@@ -0,0 +1,43 @@
...
+    local_map)
+        echo "Creating datamaps image tiles."
+        cd ..
+        exec ./docker/run_local_map.sh
+        ;;
+    shell)
+        echo "Opening shell"
+        cd ..
+        if [ -z "$*" ]; then
+            exec /bin/bash
```
willkg: Should move the `echo "Opening shell"` line to before this one.

docker/config/app.env
```diff
@@ -0,0 +1,8 @@
+# Environment variables for local dev environment
+
+REDIS_HOST=ichnaea_redis_1
+
+# Database bits
+DB_HOST=ichnaea_mysql_1
+DB_USER=root
+DB_NAME=location
```
willkg: It's likely there are other things we want to set in the local dev environment. For example, we probably want a different logging configuration--one that's not quiet as a mouse.

I'm pushing that off to another PR. Probably will do that and switch to Everett.

docker/config/app.env
```diff
@@ -0,0 +1,8 @@
+# Environment variables for local dev environment
+
+REDIS_HOST=ichnaea_redis_1
+
+# Database bits
+DB_HOST=ichnaea_mysql_1
```
willkg: These are probably wrong settings. They should be the service names which are exposed as host names inside the containers.

docker/config/test.env
```diff
@@ -0,0 +1 @@
+DB_NAME=test_location
```
willkg: This covers getting the tests to run against a different db. We might want to do the same thing for redis, but it kind of looks like all the redis usage is mocked out.

docker/mysql/Dockerfile
```diff
@@ -1,6 +1,5 @@
-FROM mysql:5.7
+FROM mysql:5.7.27
 ENV MYSQL_ROOT_PASSWORD=location
-ENV MYSQL_DATABASE=location
```
willkg: This line caused the container to create the `location` database at startup. We now have code that creates/drops databases, so we don't need this anymore.

docker/node/Dockerfile
```diff
@@ -1,9 +1,8 @@
 FROM node:8-slim
 
-# add a non-privileged user for installing and running
-# the application
-RUN groupadd -g 10001 app && \
-    useradd -d /app -g 10001 -G app -M -s /bin/sh -u 10001 app
```
willkg: The `node` image creates a `node` user (gid 1000) and `node` group (gid 1000). Since this is only run in local development environment, we don't need it to use 10001 id and don't need to create an app user. So I nixed all that.

docker/run_local_map.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python ./ichnaea/scripts/datamap.py --create --output /app/ichnaea/content/static/tiles/
```
willkg: This will run, but kicks up errors. I'll fix this in a future PR.

docker/run_map.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+python ./ichnaea/scripts/datamap.py --create --upload
```
willkg: This will run, but kicks up errors. I'll fix this in a future PR.

docker/run_web.sh
```diff
@@ -0,0 +1,15 @@
...
+PORT=${PORT:-8000}
+echo "Running webserver on http://localhost:${PORT} ..."
+
+gunicorn \
+	--pythonpath /app \
+	--error-logfile=- \
+	--access-logfile=- \
+	--log-file=- \
+	--config=python:ichnaea.webapp.settings \
+	--bind 0.0.0.0:"${PORT}" \
```
willkg: In Socorro-land, we have configuration for a few other things:

```
GUNICORN_WORKERS=${GUNICORN_WORKERS:-"1"}
GUNICORN_WORKER_CLASS=${GUNICORN_WORKER_CLASS:-"sync"}
GUNICORN_MAX_REQUESTS=${GUNICORN_MAX_REQUESTS:-"10000"}
GUNICORN_MAX_REQUESTS_JITTER=${GUNICORN_MAX_REQUESTS_JITTER:-"1000"}
```

In Ichnaea, that's in `ichnaea.webapp.settings`. I think we should move it here so it can be set by environment variables and managed along with the infra code. I'm going to push those changes off to a new PR.

docs/local_dev.rst
```diff
@@ -0,0 +1,253 @@
...
+If there were changes to the requirements files or setup scripts, you'll need to
+build new images::
+
+    $ make build
+
+
+If there were changes to the database tables, you'll need to wipe the MySQL
+database and Redis::
+
+    $ make setup
```
willkg: When we figure out the sample data story, we can amend this step accordingly.

ichnaea/config.py
```diff
@@ -20,15 +20,15 @@
 with open(CONTRIBUTE_FILE, 'r') as fd:
     CONTRIBUTE_INFO = simplejson.load(fd)
 
-VERSION = pkg_resources.get_distribution('ichnaea').version
```
willkg: Since we no longer install ichnaea, we can't pull the version from the distribution data.

ichnaea/conftest.py
```diff
@@ -28,7 +29,7 @@
 from ichnaea.config import (
     ALEMBIC_CFG,
 )
-from ichnaea.db import configure_db
+from ichnaea.db import configure_db, create_db, drop_db
```
willkg: We don't use `drop_db` in this file. We can remove that from the imports.

ichnaea/scripts/db.py
```diff
@@ -0,0 +1,53 @@
...
+def create_database():
+    """Create a database."""
+    print("Creating database %s...." % DB_DDL_URI)
+
+    try:
+        create_db(uri=DB_DDL_URI)
+    except ProgrammingError:
+        print("Database already exists.")
+        return 1
+
```
willkg: I can nix this empty line.

setup.py
```diff
@@ -2,13 +2,7 @@
 import os
```
willkg: Given that we're no longer installing ichnaea Python package, we might not need this `setup.py` file anymore.

willkg:
Testing this out:

1. `make build`
2. `make setup`
3. `make runservices`
4. `make run` and `make runcelery` in two separate windows--they should run
5. `make test` and `make lint` should run and pass
6. `make testshell` gets you a shell in the test environment where you can run `pytest`
7. `make buildjs`, `make buildcss`, and `make docs` should all run--they all change the state of the repository, so you probably want to throw away changes afterwards
8. `make mysql` gets you into a mysql prompt

I think that's all the critical stuff. We'll hone this in future PRs.
8. `make shell

Makefile
```diff
@@ -0,0 +1,126 @@
...
+shell: my.env .docker-build
+	${DC} run --rm app shell
+
+.PHONY: mysql
+mysql: my.env .docker-build
+	${DC} up -d mysql
+	@echo "=============================="
+	@echo "Root password is \"location\"."
+	@echo "=============================="
+	${DC} exec mysql mysql --user root --password
```
jwhitlock: How about adding the password and automatically connecting to the `location` database?

```
${DC} exec mysql mysql --user root --password=location location
```

docker/config/test.env
jwhitlock: When I try ``make testshell``, and then ``pytest``, the tests get stuck. The "right" way to run tests is ``TESTING=true pytest``. We could add a line to this file so that just pytest works:

```
TESTING=true
```

``make shell`` doesn't load this file, so ``TESTING`` will remain unset.

docker/run_tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# Run tests.
+#
+# Run this in the app container.
+
+TESTING=true pytest "$@"
```
jwhitlock: Even if we move ``TESTING=true`` to ``test.env``, I think it is useful to manually set it here.

docker/run_worker.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# BBB: Celery 4
```
jwhitlock: Do you know what BBB stands for? It seems to mean "TODO", and always is used around Celery 4 code.

docs/local_dev.rst
```diff
@@ -0,0 +1,253 @@
...
+You can run the test suite like this::
+
+    $ make test
+
+
+If you want to pass different arguments to pytest or specify specific
+tests to run, open up a test shell first::
+
+    $ make testshell
+    app@blahblahblah:/app$ pytest [ARGS]
```
jwhitlock: This needs a ``TESTING=true``, unless we add to ``test.dev``.

Makefile
willkg: Aha! I couldn't get `mysql --user root --password location` working--seemed like `--password` was treated as an indicator to ask for the password so I did this janky thing.

But it works with `--password=location` and that's awesome! I'll fix it.

docker/config/test.env
willkg: Oops--that's a mistake. Good catch!

Also, we should probably do a pass on removing/reducing uses of `TESTING` so as to reduce complexity.

docker/run_worker.sh
willkg: I have no idea. I don't think I've seen BBB in code before.

@hannosch ^^^ What does BBB mean?

willkg:
Thank you!

docker/run_worker.sh
hannosch: BBB means “backwards compatibility code”, so things which can be dropped or changed once a dependency upgrade is done or some old Python version no longer needs to be supported. It is used in open-source projects, which also use XXX as a means to signal an urgent todo task / something that must be fixed before the next release. Other projects use TODO, FIXME or other markers.

The Celery 4 stuff is around, as I did the upgrade of all the code to Celery 4 including adjusting to all the changed config settings. But than had to roll back to Celery 3 as version 4 wasn’t production ready at the time. Since these changes were all over the place, I kept them in the code, so it would have been easy to do the upgrade later on. I did expect that to happen a couple weeks later, not years after.

PR #844, opened 2019-08-15
bug 1573951: Terms of service, other text changes

jwhitlock:
* Add Developer Terms of Service: Mozilla Location Service Query API at `/terms`, along with some page-specific styles.
* Add new form-based method for requesting an API key.
* Reword statement about omitting WiFi data from downloads.
* Drop reference to the leaderboard from the privacy page.
* Drop IRC and email as contact methods, leaving the public mailing list.

It took a bit to get the 1.5.x branch running locally, following the [development instructions](https://jwhitlock.github.io/ichnaea/install/devel.html). I went through a few cycles of running `make` and installing libraries.

I needed to wipe out my MySQL database, which was based on the master branch (``docker-compose stop --volumes``), and then running `make` again to initialize a fresh DB.

The asset minimizing system is not working (a `bower` update is required for the new repository URL), so these styles are manually copied (non-minimized) to `bundle-base.css`.

I copied ``location.ini`` to ``location-mine.ini``, and added a section to enable the website:

```
[web]
enabled = true
```

My command to run it was:

```
ICHNAEA_CFG=location-mine.ini bin/gunicorn -b 127.0.0.1:7001 \
    -c python:ichnaea.webapp.settings ichnaea.webapp.app:wsgi_app
```

I was then able to see the web pages on http://127.0.0.1:7001. Template changes require restarting the service, ``bundle-base.css`` changes a browser refresh.

If you don't want to go through this process to see the changes, I can post screenshots, such as the new terms page:

Screenshot_2019-08-15 MLS - Developer Terms of Service Mozilla Location Service Query API


willkg:
I <3 that you did that first commit. Thank you!

ichnaea/content/templates/terms.pt
```diff
@@ -0,0 +1,204 @@
...
+            <span class="legal-term">Data Collection</span> -
+            We adhere to the practices outlined in the
+            <a href="https://www.mozilla.org/privacy/">Mozilla privacy policy</a>
+            for how we receive, handle and share information we collect
+            providing MLS. The following data is required for use of our
+            service: Name, Email, Organization (as part of registration). To
+            determine location, end user devices send us a combination of IP
+            address and nearby public Wi-fi networks and cell phone towers,
+            which we use to provide an approximate location. We do not keep
+            this data or track users. Mozilla retains aggregate data of the
```
hannosch: This isn’t quite right. All user queries are stored and used to revalidate the available data. This data isn’t associated with any form of user identifier, but the data itself is stored and used.

jwhitlock:
Force-push to remove extra query string parameter (edit requested) from the the google form URL.

ichnaea/content/templates/terms.pt
ckolos: The queries we're talking about here.... are they simply used to check the DB to see if we have that point already, bumping last_seen if so and adding it if not?

hannosch: The last part of https://mozilla.github.io/ichnaea/install/architecture.html#web-service should be correct. The data is used to validate and change the position estimate of existing data. It also adds new unknown networks found in the query. In order to do this, all queries get stored in Redis for a bit. Whether or not those also get dumped into an S3 bucket for long term storage ... I don't remember.

That long term storage decision is a policy encoded in the ``export_config`` database table and not in code. In it are a set of export targets, like s3 and potential commercial partners. That table also has a set of exclusions for both API keys and sources of data. But I don't remember what those exclusions looked like anymore.

jwhitlock: Thanks @hannosch! This looks like the relevant bit from the architecture docs:

> Should the service be able to generate a plausible and good enough answer, this is send back as a response. The incoming query and this answer is also stored in another queue, to be picked up by the data pipeline later. This query based data is used to in/validate the database contents and estimate the position of previously unknown networks as to be near the already known networks.

I'll get familiar with the geolocate and fallback code, so I can make an informed technical recommendation on updating this text.

PR #849, opened 2019-08-20
Update docker tagging for "make build" target

jwhitlock:
``make build`` now tags the built image as ~``ichnaea_node:latest``~ ``local/ichnaea_app`` rather than ``mozilla/location``. Fix the circleci build step that that re-tags and pushes to dockerhub.

Follow-on work for issue #828.

**Update**: Force push to fix local tag name, commit comment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,12 +97,12 @@ jobs:
               # push on master or git tag
               if [ "${CIRCLE_BRANCH}" == "master" ] || [ -n "${CIRCLE_TAG}" ]; then
                 echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
-                docker tag "mozilla/location" "mozilla/location:${DOCKER_TAG}"
+                docker tag "local/ichnaea_app" "mozilla/location:${DOCKER_TAG}"
                 retry docker push "mozilla/location:${DOCKER_TAG}"
 
                 # push `latest` on master only
                 if [ "${CIRCLE_BRANCH}" == "master" ]; then
-                  docker tag "mozilla/location" "mozilla/location:latest"
+                  docker tag "local/ichnaea_app" "mozilla/location:latest"
                   retry docker push "mozilla/location:latest"
                 fi
               fi


.circleci/config.yml
```diff
@@ -97,12 +97,12 @@ jobs:
               # push on master or git tag
               if [ "${CIRCLE_BRANCH}" == "master" ] || [ -n "${CIRCLE_TAG}" ]; then
                 echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
-                docker tag "mozilla/location" "mozilla/location:${DOCKER_TAG}"
+                docker tag "ichnaea_node:latest" "mozilla/location:${DOCKER_TAG}"
```
willkg: `ichnaea_node` is a node.js container that builds static assets. I think you want `local/ichnaea_app`:

```
saturn5 ~/mozilla/ichnaea> docker image ls | grep ichnaea
(M=dd51d) saturn5 ~/mozilla/ichnaea> docker image ls | grep ichnaea
ichnaea_mysql                                       latest               79a670413fb0        2 minutes ago       373MB
ichnaea_node                                        latest               bf57fa1aeba7        3 minutes ago       427MB
local/ichnaea_app                                   latest               40b68da9af30        4 minutes ago       959MB
```

jwhitlock: Nice catch! I was looking at the last line of ``make build``, I should have read higher.

jwhitlock:
Force push to fix local tag name (``local/ichnaea_app``), commit comment.

jwhitlock:
> I think this is right, but I think the only way to find out is to land it.

I'm OK verifying this works after it merges to master.

In the future, if one of us pushes the branch to ``mozilla/ichnaea``, then the CircleCI build gets dockerhub credentials, and will push the images generated by the PR, tagged with the commit hash. This can be used to test changes that may impact docker image pushes. At least I hope it does, I don't think I've tried it yet.

When the branch is from a fork, dockerhub credentials are not sent to the build, and the docker  push step aborts early. This prevents a random GitHub user from changing a build step to print the dockerhub credentials.

PR #858, opened 2019-09-06
redo configuration with Everett

### LINKED ISSUE #836: switch to everett for configuration
Configuration in the master branch is defined in `ichnaea/webapp/config.py` using a bunch of `os.environ.get()` lines.

We should switch to a configuration library--that'd be easier. If we switch to everett, then we get free documentation.

This covers doing that.

willkg:
This redoes configuration across the application:

* changes some module names because "configuration" is ambiguous
* changes how settings are used so as to easily differentiate between settings and Python module "constants"
* moves non-settings out of the settings file
* collapse the DB settings into three non-composed DB URI settings
* remove some uses of the TESTING variable--we have a `test.env` file for testing-related settings
* update configuration documentation
* remove some monkey patching in the tests in favor of Everett's `config_override`

To test:

1. run `make build` then `make setup` to suss out issues in building and setting up services
2. run `make run` and go to `http://location:8000` and click around
3. run `make celery` and wait a while to see if it kicks up any errors

Fixes #836


ichnaea/alembic/versions/000000000000_base.py
```diff
@@ -13,10 +13,7 @@
 import sqlalchemy as sa
 from sqlalchemy import text
 
-from ichnaea.config import (
-    DB_RO_URI,
-    DB_RW_URI,
-)
+from ichnaea.conf import settings
```
willkg: This reminds me of Django. I don't know if there are better models for this. I like `settings` because that's exclusively used in Ichnaea to mean "settings". I put it in `ichnaea.conf` because I had to put it somewhere. I'm game for other ideas.

ichnaea/async/tests.py
```diff
@@ -10,38 +9,85 @@
 
 class TestBeat(object):
 
-    def test_tasks(self, celery):
-        tmpdir = tempfile.mkdtemp()
-        filename = os.path.join(tmpdir, 'celerybeat-schedule')
+    def test_tasks(self, celery, tmpdir):
```
willkg: This test was failing, so I spent some time tinkering with it and ended up tweaking it a bit.

Here, I switched it to use the pytest `tmpdir` fixture so we don't have to do our own tmpdir maintenance work.

ichnaea/async/tests.py
```diff
@@ -10,38 +9,85 @@
...
+            all_tasks - registered_tasks
+            == set([
+                'data.update_blue',
+                'data.update_cell',
+                'data.update_wifi',
+                'data.update_datamap',
+                'data.cleanup_datamap',
+                'data.cell_export_diff',
+                'data.cell_export_full',
+                'data.export_reports',
```
willkg: For some reason, this PR changes the output of this specific check. Before `data.update_statregion` would be in `all_tasks`, but not `registered_tasks`. Now it's in both sets.

I spent some time trying to figure out what changed and why this test broke, but I'm completely puzzled as to why it shouldn't show up in `registered_tasks` before. My current theory is that it's a bug in the test and `data.update_statregion` should have been in `registered_tasks` and something got fixed when I did this PR.

ichnaea/async/tests.py
```diff
@@ -10,38 +9,85 @@
...
+                'data.update_wifi_6',
+                'data.update_wifi_7',
+                'data.update_wifi_8',
+                'data.update_wifi_9',
+                'data.update_wifi_a',
+                'data.update_wifi_b',
+                'data.update_wifi_c',
+                'data.update_wifi_d',
+                'data.update_wifi_e',
+                'data.update_wifi_f',
```
willkg: Before, `registered_tasks` would group the tasks dropping the right-most part. That seems curious to me--seems like we really want to compare against the complete list of tasks.

ichnaea/cache.py
```diff
@@ -8,15 +8,17 @@
 import redis
 from redis.exceptions import RedisError
 
-from ichnaea.config import REDIS_URI
+from ichnaea.conf import settings
 
 
-def configure_redis(cache_url=REDIS_URI, _client=None):
+def configure_redis(cache_url=None, _client=None):
```
willkg: Because `REDIS_URI` is a setting, we don't want the value at the time this module was loaded codified as an argument default. That'll mess up the tests.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+
+    required_config.add_option(
+        'asset_bucket',
+        default='',
+        doc='???'
+    )
+    required_config.add_option(
+        'asset_url',
+        default='',
+        doc='???'
```
willkg: I should fix this doc.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+        'testing',
+        default='false',
+        parser=bool,
+        doc='Whether or not we are running tests.'
+    )
+
+    required_config.add_option(
+        'asset_bucket',
+        default='',
+        doc='???'
```
willkg: I should fix this doc.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+        'asset_url',
+        default='',
+        doc='???'
+    )
+
+    # Database related settings
+    required_config.add_option(
+        'db_library',
+        default='pymysql',
+        doc='one of "pymysql" or "mysqlconnector"'
```
willkg: Ichnaea uses one in the webapp because it's gevent friendly and the other in other cases. I think unless there are compelling reasons to do otherwise, we should just standardize on one and not switch back and forth. That'll simplify a few things.

I have vague memories of this coming up before, but I forget whether we decided anything.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    required_config.add_option(
+        'db_readonly_uri',
```
willkg: "ro" and "rw" seemed too short and I know I read it wrong once or twice when skimming the code. For that reason, I expanded it to "readonly" and "readwrite".

I'm not sold on those settings names. I'm game for something that's clearer.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+        default=os.path.join(HERE, 'tests/data/GeoIP2-City-Test.mmdb'),
+        doc='absolute path to mmdb file for geoip lookups'
+    )
+
+    def __init__(self, config):
+        self.raw_config = config
+        self.config = config.with_options(self)
+
+    def __call__(self, *args, **kwargs):
+        return self.config(*args, **kwargs)
```
willkg: I thought about doing getattr (or whatever it is) instead and supporting uppercase names, so then we could do something like this:

```
settings.TESTING
```

instead of this:

```
settings('testing')
```

If we do that, we'll break the getattr protocol since Everett will want to throw a different exception. That might be fine, though.

@jwhitlock Does the `settings.TESTING` syntax look better to you?

ichnaea/conftest.py
```diff
@@ -379,7 +377,6 @@ def region_searcher(data_queues, geoip_db,
 def map_config(monkeysession):
     tiles_url = 'http://127.0.0.1:9/static/tiles/{z}/{x}/{y}.png'
     monkeysession.setattr('ichnaea.content.views.MAP_TILES_URL', tiles_url)
-    monkeysession.setattr('ichnaea.content.views.MAP_TOKEN', 'pk.123456')
```
willkg: This is a setting, so we can use Everett's `config_override` to set it.

ichnaea/db.py
```diff
@@ -19,29 +21,35 @@
...
+from ichnaea.conf import settings
+
 
 DB_TYPE = {
-    None: DB_DDL_URI,
-    'ddl': DB_DDL_URI,
-    'ro': DB_RO_URI,
-    'rw': DB_RW_URI,
+    # Default to readonly if db_type isn't specified
+    None: settings('db_readonly_uri'),
```
willkg: The default used to be the ddl uri, but that seems like a really bad idea to me, so I changed it to the readonly uri.

ichnaea/geoip.py
```diff
@@ -164,7 +164,7 @@
 }
 
 
-def configure_geoip(filename=GEOIP_PATH, mode=MODE_AUTO,
+def configure_geoip(filename=None, mode=MODE_AUTO,
```
willkg: `GEOIP_PATH` is a setting and we don't want the value at the time this module is loaded to persist as an argument default.

ichnaea/scripts/datamap.py
```diff
@@ -391,7 +389,7 @@ def main(argv, _raven_client=None, _stats_client=None, _bucketname=None):
 
         bucketname = _bucketname
         if not _bucketname:  # pragma: no cover
-            bucketname = ASSET_BUCKET
+            bucketname = settings('ASSET_BUCKET')
```
willkg: Oops--this should get fixed. Wrong case.

setup.cfg
```diff
@@ -2,7 +2,10 @@
 universal=0
 
 [flake8]
-ignore = N811
+ignore =
+    N811,
+    # W503: line break before operator; this doesn't work with black
+    W503
```
willkg: We're not using black, yet, but we will at some point.

willkg:
@jwhitlock Can you skim this for sanity and ergonomics?

ichnaea/async/settings.py
```diff
@@ -86,6 +83,4 @@
 CELERY_TASK_SERIALIZER = 'internal_json'
 
 # cleanup
-del REDIS_URI
 del TASK_QUEUES
```
jwhitlock: I'm not sure why there is a variable cleanup section at all. Is it for memory management? Or should these be underscore names like `_TASK_QUEUES` so Python treats them as internal variables?

ichnaea/async/tests.py
jwhitlock: My guess is that this was to minimize the test changes when these tasks exploded into buckets. I like the more explicit list.

ichnaea/conf.py
jwhitlock: My memories are also vague. It may be tied up in using a connection pool, which is good for read-only webapp connections, and a direct connection, which is better for DDL changes. Or, I may be remembering past struggles with schema changes and SQLAlchemy.

There was quite a bit of drama around Python MySQL libraries, including the Python 2 to 3 transition. I think this has mostly shaked out by 2019. I'm inclined to standardize on ``mysqlclient``, the [Django recommendation](https://docs.djangoproject.com/en/2.2/ref/databases/#mysql-db-api-drivers). I'm also inclined to investigate a switch to PostgreSQL. All of this is out of scope for this PR though.

ichnaea/conf.py
```diff
@@ -0,0 +1,117 @@
...
+        'redis_uri',
+        doc='uri for Redis; redis://HOST:PORT/DB'
+    )
+
+    required_config.add_option(
+        'mapbox_token',
+        default='',
+        doc=(
+            'Mapbox API key; if you do not provide this, then parts of '
+            'the site will not work'
```
jwhitlock: For today, this is the right code, and good job including the significance of leaving it at the default. 👍

I'd like to revisit this soon. When unset, the website is a placeholder, and when set it looks like https://location.services.mozilla.com/. I'd rather always see something that looks like the location website, with a message like "map disabled, no Mapbox API key specified", to ease development. We'll see if anyone comes forward as users of [ichnaea as a platform](https://groups.google.com/forum/#!topic/mozilla.dev.geolocation/pWB4LQ3UyrM).

ichnaea/conf.py
jwhitlock: As a Django user, ``settings.TESTING`` feels more familiar, but the syntax ``settings('testing')`` helps orient me that I'm not in Django, so let's keep it.

ichnaea/content/views.py
```diff
@@ -64,13 +60,13 @@ def configure_tiles_url(asset_url):
     return (tiles_src, tiles_url)
 
 
-MAP_TILES_SRC, MAP_TILES_URL = configure_tiles_url(ASSET_URL)
+MAP_TILES_SRC, MAP_TILES_URL = configure_tiles_url(settings('asset_url'))
```
jwhitlock: I guess an advantage of Django-style configuration is that you can calculate these derived settings in one place, and use the same test override tools for original and derived settings. It is a little ugly to test ``MAP_TILES_URL``, but it works. 👍

ichnaea/db.py
jwhitlock: My guess would be a failure in a setup or DB migration script rather than app code, which may include IT-managed scripts that aren't in this repo.

I think the chance of this breaking something is a lot less in the dockerized environments, and I like that it breaks safely rather than unsafely.

ichnaea/db.py
```diff
@@ -268,14 +276,16 @@ def create_db(type_=None, uri=None):
         "CREATE DATABASE {} CHARACTER SET = 'utf8'".format(db_to_create)
     )
 
+    alembic_cfg = get_alembic_config()
```
jwhitlock: For what it is worth, I'm more comfortable with this way of calculating derived settings than the ``MAP_TILES_URL`` setting, seems to be more mockable. But, they both work. 👍

ichnaea/async/settings.py
willkg: It's for namespace management. That way no one can import `REDIS_URI` from this module.

You suggest prefixing it with `_` so then there's the indication it's private to the module to other developers. That'd work. We could also do `__` (two underscores) and Python will mange the actual name which makes it harder to import.

I think if it were me, I'd go the "we could do it, but meh" route and just remove the cleanup section altogether.

ichnaea/async/tests.py
willkg: That's interesting. I bet you're right.

ichnaea/conf.py
willkg: I pushed off the "let's switch to postgres" to another day. I like the idea of revisiting the mysql client library, so I wrote up #870 for that.

ichnaea/conf.py
willkg: I like that idea, too. Seems weird to shut off the entire website if there's no mapbox api key plus it'd make it easier to develop if it didn't do that.

ichnaea/content/views.py
willkg: We can do derived settings using `AppConfig`, too. I'll switch that.

ichnaea/async/tests.py
jwhitlock: It looks like on master these tasks are not registered, which are now registered with this PR:

```
data.cleanup_datamap_ne
data.cleanup_datamap_nw
data.cleanup_datamap_se
data.cleanup_datamap_sw
data.update_datamap_ne
data.update_datamap_nw
data.update_datamap_se
data.update_datamap_sw
data.update_statregion
```

These all have decorators with the attribute ``_enabled=_web_content_enabled``, so in master that is ``False`` and in this branch it is ``True`` when the task decorators are evaluated. ``data.update_statregion`` is special because it doesn't have a ``_shard_model``, so the name is the same after registration.

This test is checking:
* Does the `_enabled` parameter to ``@celery_app.task`` work, skipping registration when ``False``? It looks like one value flipped from ``False`` to ``True``, breaking the test in a non-obvious way.
* Does the `shard_model` parameter to `@celery_app.task` work, creating many tasks that vary by a suffix?
* Did we get the expected all and registered tasks, but not by looking directly, but by comparing differences of sets?

This test seems problematic to me. Maybe it just makes sense to test ``all_tasks`` and ``registered_tasks`` directly, rather than try to test the set differences and loop through looking for shard suffixes. Or, it should setup the environment to ensure what tests are enabled, or should vary the expected value by ``_web_content_enabled`` and ``_cell_export_enabled``. Or maybe this shouldn't be a unit test.

willkg: Aha! That's fantastic! Thank you for figuring that out.

I think I'm going to tackle #871 which changes the whole "web content enabled/disabled" thing. I'll leave this for now and address it then.

ichnaea/content/views.py
willkg: I went to do that and then discovered that `MAP_TILES_URL` is more complicated than I thought it was. I'll look at addressing this when I look at #871 next.

ichnaea/async/settings.py
willkg: I'm going with the "we don't need this" and removing it.

ichnaea/db.py
```diff
@@ -19,29 +21,33 @@
...
-    'ro': DB_RO_URI,
-    'rw': DB_RW_URI,
+    'ddl': settings('db_ddl_uri'),
+    'ro': settings('db_readonly_uri'),
+    'rw': settings('db_readwrite_uri'),
 }
 
 DB_TRANSPORTS = {
-    'default': DB_LIBRARY,
+    'default': settings('db_library'),
```
willkg: Note that this gets set at module-load time and won't update if we do a `config_override` in the tests.

willkg:
Thank you!

PR #883, opened 2019-09-10
switch to readthedocs and update sphinx

### LINKED ISSUE #859: redo how docs get generated
With all the other projects we maintain and have maintained, when we land changes in GitHub, that triggers a webhook and the project documentation gets rebuilt on ReadTheDocs.

With Ichnaea, there's a git submodule in `docs/` with the compiled documentation. In order to update the docs on github.io, one has to compile the docs and then check them into the submodule and push that. It's very manual, has mechanisms I'm not fond of, and is different than all the other projects we (@jwhitlock and I) maintain.

This issue covers changing how docs get generated such that:

1. it doesn't use git submodules
2. docs get built automatically when PRs get merged to master

I have a strong preference for switching to ReadTheDocs.

### LINKED ISSUE #879: Update Sphinx and related modules
Update Sphinx and the related modules, used to generate the documentation.

The update from Sphinx 1.6 to 2.2 includes a change in the way contributed packages are included, and may offer an opportunity to simplify the documentation configuration.

May be related to #859.

willkg:
This switches docs building/hosting to ReadTheDocs instead of github pages. It also updates Sphinx-related dependencies to recent versions.

ReadTheDocs is already set up with a webhook. After this lands, it should build and everything should be super duper. If not, I'll follow-up and fix issues as they come up.

We'll add a redirect link to the github pages later.

Fixes #859 (switch to readthedocs)
Fixes #879 (update sphinx)


.readthedocs.yml
```diff
@@ -0,0 +1,15 @@
...
+  configuration: docs/conf.py
+
+# Optionally build your docs in additional formats such as PDF and ePub
+formats: all
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.6
+  install:
+    - requirements: docs/requirements.txt
```
willkg: We have several libraries that are written in C. To make it easier to build the docs on RTD, I figured I'd limit the number of additional things we install to the smallest number possible. That means there's a tiny bit of duplication between this file and `requirements/*` files.

If we don't update this one, it's no big deal. At worst, the docs won't build, someone will get an email, and we'll fix it. I think that's a better situation than trying to eliminate duplication across requirements files so we can have one set for all things. I predict that'll be harder to do.

docs/conf.py
```diff
@@ -1,9 +1,17 @@
...
 
+import os
 import sphinx_rtd_theme
+import sys
 
-project = u'Ichnaea'
-copyright = u'2013-2019, Mozilla'
+
+REPO_DIR = os.path.dirname(os.path.dirname(__file__))
+print(REPO_DIR)
```
willkg: Oops--errant print. I'll remove this.

docs/local_dev.rst
```diff
@@ -44,14 +44,6 @@ Setup Quickstart
...
    <https://github.com/mozilla/ichnaea>`_.
 
-   Ichnaea publishes documentation at `<https://mozilla.github.io/ichnaea>`_
-   which is a git submodule in the repository.
-
-   After you clone the repository, you need to do this in the repository
-   directory::
-
-       $ git submodule update --init --recursive
-
```
willkg: Yay!

willkg:
self-r+ since this is docs and I need it to land before I know if it's working anyhow.

PR #885, opened 2019-09-10
Remove the built docs submodule

jwhitlock:
Remove the built HTML documents submodule. Documents can still be built locally with `make docs`, but they will be ignored by git.

Part of #859.


willkg:
Thank you for doing this!

PR #889, opened 2019-09-11
paint it black

### LINKED ISSUE #843: switch to black for python code formatting
black is a Python code formatter: https://black.readthedocs.io/en/stable/

We switched to black in Socorro and Tecken because it's in use by other projects. Switching to using black is not hard. When I switched Socorro, the hardest part was reviewing the PRs and verifying the changes. We can reuse some of the scaffolding we wrote there.

We should switch ichnaea to use black.

willkg:
This does a number of things:

* adds black as a dependency
* adjusts the `make lint` rule to also run `black --check`
* adjusts the flake8 configuration to match black
* reformats all the Python files
* fixes a few issues I discovered while reformatting
* adds `make lint` to CI

I broke up the changes into a number of commits. It's still probably far too hard to look at in the GitHub web ui.

To test, I did the following:

1. I reviewed all the code changes looking for instances of "wow--that's a weird change" which usually indicate other issues
2. I ran `make build`, `make setup`, `make test`, `make lint`, `make lintfix`, `make run`, and `make runcelery` looking for errors

My experiences reformatting antenna and socorro were pretty uninteresting. In both cases, we reformatted the code and tweaked a couple of weird changes, but everything was fine. I expect ichnaea to be similar.

Fixes #843


willkg:
self-r+.

PR #896, opened 2019-09-12
Add gunicorn settings

### LINKED ISSUE #891: convert gunicorn settings to environment variables
There are settings for running the webapp with gunicorn defined in `ichnaea/webapp/gunicorn_settings.py`. This list is missing a few things we want to configure. Further, none of those get values from the environment.

This issue covers adding the missing variables that we'll want to tweak and also changing them so they're set from the environment with decent defaults.

willkg:
This reworks the gunicorn setup exposing many of the settings as
environment variables and adding it to the documentation.

Fixes #891

### DIFF
--- a/docker/run_web.sh
+++ b/docker/run_web.sh
@@ -4,9 +4,7 @@ set -e
 
 # Runs the webapp in gunicorn.
 
-# NOTE(willkg): The docs literalinclude part of this file, so if you change
-# this file, you need to update the line numbers specified in
-# docs/install/config.rst.
+# START GUNICORN CONFIGURATION
 
 # Port for gunicorn to listen on
 GUNICORN_PORT=${GUNICORN_PORT:-"8000"}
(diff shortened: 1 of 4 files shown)


willkg:
Oops--to test this, do:

```
$ make run
```

It'll spit out gunicorn things. The thing it still doesn't do is output anything to the log when you go to `http://localhost:8000` in your browser. I think the logging isn't right. I wrote up #897 for that.

jwhitlock:
> I kind of wish there was a way to do a literal include of all the lines in a file between a set of BEGIN/END markers. I don't see a way to do that or an existing plugin, so I'm tossing around writing a plugin to do it. Seems handy to have.

Do the ``start-after`` and ``end-before`` options of [literalinclude](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude) do what you want?

willkg:
Rrrr... It's so annoying when documentation buries things in paragraphs. I totally missed the `start-after` and `end-before`. I'll switch to that. Thank you!

willkg:
Thank you!

PR #900, opened 2019-09-13
Add bits to run webapp without gunicorn

### LINKED ISSUE #898: document how to run webapp without gunicorn
If gunicorn is failing to start, it may fail to start with absolutely no helpful information. Since we run with gevent, using `--preload` doesn't really work. One has to switch the worker class to sync and then add preload. Even then, gunicorn may not tell you anything.

With Django projects, we can use the shell and run `./manage.py runserver` which will give us tracebacks and make a class of problems much easier to debug.

Pyramid projects are WSGI apps, so we should be able to do something similar.

This issue covers implementing whatever is needed and then documenting it in a trouble-shooting section in the docs somewhere.

willkg:
This lets us run the webapp without gunicorn. It makes some kinds of problems easier to debug.

Fixes #898


docs/install/debug.rst
```diff
@@ -4,15 +4,50 @@
 Debugging
 =========
 
-.. Note:: 2019-08-16: This needs to be updated.
+Debugging webapp
+================
```
jwhitlock: This looks like two `<h1>` sections with no content between - do we need to adjust heading levels?

willkg: The first one has over and under and is the header for the entire page, but it's hard to see because the previous line is hidden.

willkg:
Thank you!

PR #901, opened 2019-09-13
Update gunicorn, gevent, and greenlet

### LINKED ISSUE #873: Update gunicorn, related modules
Update gunicorn and related modules gevent and greenlet

willkg:
This updates the gunicorn, gevent, and greenlet dependencies.

This also removes gevent monkey patching from the tests--it wasn't
getting run early enough so it was emitting a warning and the comment
suggested it should only monkey patch the webapp tests. We can do that,
but if we do do that, we need to separate the webapp tests into a
separate test suite. Until we do that, I think it's best to remove the
monkey patching altogether.

Gunicorn 19.7.1 to 19.9.0:

* https://docs.gunicorn.org/en/stable/news.html
* I don't see anything interesting there.

gevent 1.2.2 -> 1.4.0: 

* https://github.com/gevent/gevent/blob/master/CHANGES.rst
* I didn't see anything interesting.

greenlet 0.4.12 -> 0.4.15:

* https://github.com/python-greenlet/greenlet/blob/master/NEWS
* The NEWS file is bereft of information.

Fixes #873


ichnaea/conftest.py
```diff
@@ -82,14 +82,6 @@ def monkeysession(request):
 
 @pytest.fixture(scope="session", autouse=True)
 def package():
-    # We do this here as early as possible in tests.
-    # We only do it in tests, as the real celery processes should
-    # run without the monkey patches applied. The gunicorn arbiter
-    # patches its worker processes itself.
-    from gevent import monkey
-
-    monkey.patch_all()
```
willkg: After upgrading gunicorn and gevent, this kicks up an error due to a warning gevent added where it checks a bunch of modules to see if they were loaded before monkey patching happened.

I took this out for a couple of reasons. First, it seems to suggest we can just monkey patch the webapp tests, but that doesn't fly--it patches everything for everything. Second, it kicks up the error.

If we really want to run the webapp tests monkey patched, they should be a completely separate test suite that we run separately that monkey patches earlier in the test run. Doing what we're doing now has undefined behavior.

willkg:
To test:

1. run the tests--they pass!
2. run `make run` which runs gunicorn with the gevent worker and tool around on the website

willkg:
Thank you!

PR #910, opened 2019-09-18
Update to flake8 3.7.8

jwhitlock:
* flake8 3.4.1 → 3.7.8: Update tool version, fix plugin bugs (see [Release Notes](http://flake8.pycqa.org/en/latest/release-notes/index.html))
* pycodestyle 2.3.1 → 2.5.0: add keywords async and await, detect invalid escape sequences, other formating issue detection (see [changes](https://pycodestyle.readthedocs.io/en/latest/developer.html#changes)).
* pyflakes 1.5.0 → 2.1.1: Drop support for EOL Python versions, tune underscore _ variable assignment rules, other new rules (see [news](https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst))
* entrypoints 0.3: New dependency of flake8 3.7.0. See https://github.com/takluyver/entrypoints.

These new rules are compatible with our black-formatted code, so no code change needed.

I checked if the flake8 rules needed adjusting, and I do not get linting error ``E501`` for line lengths. The other ignored rules ``E203`` and ``W503`` are still needed for black compatibility.


setup.cfg
```diff
@@ -5,8 +5,6 @@ universal=0
 ignore =
     # E203: Whitespace before ':'; doesn't work with black
     E203,
-    # E501: line too long
```
willkg: I copied this ignore set from socorro or wherever. I thought black would occasionally let lines exceed 88 characters depending on their complexity and thus quelling this warning was helpful. I'm not seeing evidence of that in the docs, though, and the line-length section is this one:

https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

Anyhow, seems like either the situation changed or I misunderstood the docs when I did this.

jwhitlock:
Ah, there it is. Black is OK with this docstring going past column 88:

https://github.com/mozilla/ichnaea/blob/e2ab34f68220dd5fd39cef34665e4d0860f11c8d/ichnaea/conf.py#L21-L25

@willkg - do we want flake8 to ignore this and use Black's line length rule, or make docstrings also fit the 88 rule?

willkg:
That's my bad--I should have fixed that docstring when I made it.

I don't think I have a preference for setting things. Whatever you want to do is fine.

jwhitlock:
Force push to add this new docstring knowledge to the commit message, and reformat the docstring to column 88.

PR #911, opened 2019-09-18
redo logging and add support for mozlog format

### LINKED ISSUE #897: fix logging in webapp
The webapp and possibly the celery scheduler and worker don't log enough in the local dev environment.

The webapp could probably use a "in dev" configuration for logging here:

https://github.com/mozilla/ichnaea/blob/a6a5426413219878e8d32152142beb2a1e2cb89e/ichnaea/log.py#L48-L53

We do similar things in Socorro and Antenna. We should do something like that here.

willkg:
This redoes how logging works in the webapp.

* adds a `LOCAL_DEV_ENV` configuration variable which lets us denote whether to set things up for local development or for a server environment
* reworks logging levels, formatters, handlers, and how all that's configured
* switches from wsgiref to waitress (pyramid uses this already) for serving the webapp in local dev mode
* adds a wsgi middleware for logging the request path, status code, and size in bytes in local dev mode

To test:

1. run `make build`
2. run `make run` -- this shows the webapp in local dev mode
3. set `LOCAL_DEV_ENV=False` in my.env and then run `make run` -- this shows the webapp in server mode with mozlog formatted logs

Fixes #897


docker/config/my.env.dist
```diff
@@ -9,3 +9,6 @@
 #
 # Set this to the Mapbox API token.
 # MAPBOX_TOKEN=
+
+# Set this if you're having serious problems that need Pyramid debugging.
+# PYRAMID_DEBUG_ALL=True
```
willkg: Figured I'd add this. It shows template data and some other things. It's probably interesting for certain scenarios.

docker/run_web.sh
```diff
@@ -35,21 +35,31 @@ GUNICORN_LOGLEVEL=${GUNICORN_LOGLEVEL:-"info"}
...
-    --log-file=- \
-    --log-level="${GUNICORN_LOGLEVEL}" \
-    --timeout="${GUNICORN_TIMEOUT}" \
-    --config=python:ichnaea.webapp.gunicorn_settings \
-    --bind 0.0.0.0:"${GUNICORN_PORT}" \
-    ichnaea.webapp.app:wsgi_app
+LOCAL_DEV_ENV=${LOCAL_DEV_ENV:-"False"}
+
+
+if [ "${LOCAL_DEV_ENV}" == "True" ]; then
```
willkg: We do this kind of thing with Socorro where if we're in local dev environment mode, we run `./manage.py rusnerver` and in server mode run with gunicorn. It makes a bunch of development a lot easier.

ichnaea/webapp/app.py
```diff
@@ -2,10 +2,16 @@
 Holds global web application state and the WSGI handler.
 """
 
-from wsgiref.simple_server import make_server
+import logging
+
+from waitress import serve
```
willkg: I swapped wsgiref out for waitress which pyramid depends on already. It's more featured and probably better to use.

ichnaea/webapp/app.py
```diff
@@ -38,7 +44,26 @@ def worker_exit(server, worker):
     shutdown_worker(_APP)
 
 
+def log_access_factory(wsgi_app):
```
willkg: This is a WSGI middleware that wraps the ichnaea webapp wsgi app. It logs data about the request and response. We can add other things to it if we like.

willkg:
Thank you!

PR #912, opened 2019-09-18
Update to geoip2 2.9.0, maxminddb 1.4.1

### LINKED ISSUE #872: Update geoip2, maxminddb modules
Update the modules that determine a region based on the maxmind DB

jwhitlock:
Update the libraries used to talk to the MaxMind API or local database. These changes do not appear to impact our usage. Fixes #872.

* geoip2 2.5.0 → 2.9.0: Drop Python 2.6, allow reading DB via file descriptor as well as path, more attributes like ``is_anonymous_vpn`` and ``is_in_european_union``.  See [history](https://github.com/maxmind/GeoIP2-python/blob/master/HISTORY.rst).
* maxminddb 1.3.0 → 1.4.1: File descriptor change, and require string for IP address (rather than ``IPv4Address`` instance). See [history](https://github.com/maxmind/MaxMind-DB-Reader-python/blob/master/HISTORY.rst).


jwhitlock:
Tests pass, using the local file ``ichnaea/tests/data/GeoIP2-City-Test.mmdb`` as set by the env variable ``geoip_path``.

Issue #874, opened 2019-09-09
Update redis module


jwhitlock:
Update redis module, used to connect to the redis database

jwhitlock:
The notes for upgrading from redis 2.x to 3.x mention that [ssl host validation is now on by default](https://github.com/andymccurdy/redis-py#ssl-connections), and that the SSL certs from ElasticCache don't always have the header set correctly. 3.0 was released last year, so maybe AWS is better, but I imagine we'll need to disable SSL host validation in production.

jwhitlock:
Celery 3 doesn't work with Redis 3.x. The signature of the ``zadd`` command changed, and the Celery 3 version of Kombu uses the Redis 2.x. We can update to the latest Redis 2.x, but the 3.x upgrade is blocked by #862, and may have to be done at the same time.

PR #914, opened 2019-09-19
Update to celery 3.1.26.post2, redis 2.10.6

### LINKED ISSUE #861: Update to latest celery 3
Update to the latest versions of Celery 3

jwhitlock:
* celery 3.1.25 → 3.1.26.post2: Fix issue when switching between celery 4.1.x and 3.1.x protocols.
* redis 2.10.5 → 2.10.6: Perf improvements, bug fixes, support new Redis commands like HSTRLEN and TOUCH. Included by ``celery[redis]`` extra, but used in the code, so kept in defaults.
* kombu 3.0.37 moved from defaults to constraints, with no version change. This module is used directly in the code, which is the standard for being in defaults, but it is tied so closely to celery that it belongs in constraints.
* pytz 2017.2 → 2019.2: Keep updating those timezones. Also, move to constraints, since it is required by celery.

Fixes #861 (again).


requirements/default.txt
```diff
@@ -87,7 +87,7 @@ gevent==1.4.0 \
 gunicorn==19.9.0 \
     --hash=sha256:aa8e0b40b4157b36a5df5e599f45c9c76d6af43845ba3b3b0efe2c70473c2471 \
     --hash=sha256:fa2662097c66f920f53f70621c6c58ca4a3c4d3434205e608e121b5b3b71f4f3
-# TODO: Update to redis[hiredis]>=2.10.6 or later, and move hiredis to constraints
+# TODO: Update to redis[hiredis]>=3.0, and move hiredis to constraints
```
jwhitlock: I misunderstood the sequence of ``setup.py`` changes, redis 3.0 is the first version that supports ``hiredis`` as an extra, and will be updated as part of #874.

The [redis 3.3.0 release notes](https://github.com/andymccurdy/redis-py/blob/d811ae71dbdbeeb4fc0ee73a96b7fbdb1aec8522/CHANGES#L38-L40) include this line:

> * Added support for hiredis-py 1.0.0 encoding error support. This should make the PythonParser and the HiredisParser behave identically when encountering encoding errors. Thanks Brian Candler.

This suggests the hiredis update should wait for at least redis 3.3.0

jwhitlock:
For local testing, I made sure tests pass (``make build test``), ran the website and clicked around (shouldn't matter for these updates) and ran celery (``make runcelery``) with a Mapbox token so that the data tasks would run without errors or warnings.

requirements/default.txt
```diff
@@ -179,22 +176,9 @@ pytest==3.2.0 \
...
-    --hash=sha256:c883c2d6670042c7bc1688645cac73dd2b03193d1f7a6847b6154e96890be06d \
-    --hash=sha256:d1d6729c85acea5423671382868627129432fba9a89ecbb248d8d1c7a9f01c67 \
-    --hash=sha256:ddc93b6d41cfb81266a27d23a79e13805d4a5521032b512643af8729041a81b4 \
-    --hash=sha256:f5c056e8f62d45ba8215e5cb8f50dfccb198b4b9fbea8500674f3443e4689589
 raven==6.1.0 \
     --hash=sha256:02cabffb173b99d860a95d4908e8b1864aad1b8452146e13fd7e212aa576a884 \
     --hash=sha256:56dc9062dd42bca97350e5048ff417c914376366caa3b1b5f788b27ddc0a34b7
-redis==2.10.5 \
-    --hash=sha256:5dfbae6acfc54edf0a7a415b99e0b21c0a3c27a7f787b292eea727b1facc5533 \
-    --hash=sha256:97156b37d7cda4e7d8658be1148c983984e1a975090ba458cc7e244025191dbd
```
willkg: There's a bunch of code that imports `redis`. Looks like it's used for caching and then a bunch of modules import it to get `RedisError`. Seems like we need to keep this in the `default.txt` file.

I don't see anything that uses hiredis. Doing a `pipdeptree` suggests it's not required by other packages.

jwhitlock: It's optional, used by the redis module if available. The explicit extra was added in redis 3.0.0. See https://github.com/mozilla/ichnaea/pull/825#discussion_r310264092.

jwhitlock: and 👍 to moving redis back. I confirmed that ``celery[redis]`` brings it in, but we can be more explicit.

willkg: Ahh... ok--that makes sense. I didn't understand https://github.com/mozilla/ichnaea/pull/825#discussion_r310264092 .

jwhitlock:
Thanks!

Issue #503, opened 2017-03-21
Import sample data


Alixpat:
Hi,

How should I do to import a sample data ?

Regards.

Manicben:
Regarding this issue, would it be possible to get an explanation of the MySQL database table columns in the meantime?

I've been loading CSVs manually from https://www.mylnikov.org/download into my own hosted Ichnaea instance. In order to make this work I ensured that the columns matched, however I didn't understand why there was a separate 'cellid' and 'cid' column. Afterwards I found out that 'cellid' was a binary field and due to being the primary key, served as some kind of lookup key during request queries.

None of this is covered in the current documentation on Ichnaea, so I was hoping to get a response here if possible. Of course, adding some method of importing data (in CSV format) to Ichnaea would be very helpful in future too.

hannosch:
Currently the 2.x version has no import script, it will get added back in the future again. If one needs to load cell data into this version, the old script from the 1.5 branch (https://github.com/mozilla/ichnaea/commit/5d58e93eedcd25f12b457709b888498af790ec25#diff-a223b4aea91068919bd1660253a87b0d) might help to explain the mapping of CSV fields to database columns.

The most important part is constructing the right primary key, as a binary field combining the radio, mcc, mnc, lac and cid values into one cellid column. The encode_cellid function and the CELLID_STRUCT (https://github.com/mozilla/ichnaea/blob/master/ichnaea/models/cell.py#L296) should give hints as to how to do that.

willkg:
I'm adjusting the scope of this to "write tools for generating sample data for a local dev environment". I'm not sure what other use cases there are for sample data other than development. If there are other use cases, we can defer those to new issues.

My goal is to build enough sample data that we can test and debug the major features of ichnaea and write a basic system test:

* locate lookup api
* region lookup api
* geosubmit api
* internal export
* public cell export to the downloads page

There's a ton of details in those APIs, so I think I'll do a basic set of sample data for this issue that covers some things and we can improve that via future issues and pull requests as needs come up.

I'm going to add a cli to add/manipulate data in the `api_key` table. That seems pretty critical to testing API endpoints.

The base migration creates an "internal" export in the `export_config` table which covers the internal export. After we add an s3-equivalent, we should add a "bucket" export, too. At some point, maybe a "geosubmit" export, but that probably requires a dummy ichnaea service. Some day.

I think that leaves these populating these tables with data:

* blue_shard_N
* cell_area
* cell_gsm
* cell_lte
[...]

Issue #890, opened 2019-09-11
automatically expire keys


willkg:
In MLS, we have a different types of API keys and one of those types needs to auto-expire in 3 months. Currently, there's no way to do that.

This issue covers adjusting the api key table to have some additional bookkeeping information in it like an "expires_on" field and add a scheduled task to go through and expire api keys that should be expired.

willkg:
I'm moving this to Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1600845

Issue #846, opened 2019-08-19
add an admin


willkg:
Currently, the way we manage API keys and investigate data in the db is manually on an admin node.

We're in the process of redoing the infrastructure and won't have access to admin nodes.

This issue covers looking into building out a web admin. Then we'll write up an issue for API key management requirements.

hannosch:
The need for an admin web UI was never urgent enough, as I was the only real website admin and had production access. So it was easier to just pre-create 100 api keys, insert them into the database and than do all management in a Google spreadsheet. Totally low-key, but good enough at the time.

willkg:
One thing we need to be wary of is that the webapp uses the ro db uri--it's read-only. The admin will need to be read/write.

It might make more sense to have the admin be a separate webapp that runs on a separate node on a different domain.

willkg:
I tweaked the title to generalize a bit.

The new admin will be tied to the new infrastructure. However, if it turns out we need something sooner that works on the current infrastructure, I can throw together a one-file Python script cli.

jwhitlock:
I created a duplicate, now closed, with this description:

Create a website for administering a Ichnaea instance. The first version would be restricted to maintaining the configuration tables ``api_key`` and ``export_config``. In Mozilla's production environment, it should use single-sign on and be restricted to core team members. It may use a known username and password in the development environment.

If I was creating it, I would use the Django admin, and create parallel Django models for the SQLAlchemy models for these two tables.

First proposed as part of [bug 1600845 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1600845#c4)

PR #915, opened 2019-09-19
Implement basic apikey manipulation cli

willkg:
This implements a basic apikey manipulation cli that lets you list keys, show key details, and create keys.

In doing that, I removed the code in the base migration that creates database users and a test key. Seems like it's safer if both of them are created explicitly with intention rather than accidentally by running db migrations.

This is part of issue #503. I figured I'd break it up into parts as I finished them rather than one big sample data code dump.

To test:

1. `make build`
2. `make setup`
3. `make test` to verify tests still work
4. `docker-compose run app shell ./ichnaea/scripts/apikey.py` and play with that

That `docker-compose` line is a bit much. With Socorro, we wrote a `socorro-cmd` as a frontend for all the scripts. I removed the `dev` script from this repo that sort of served a similar purpose a month ago. Maybe we should re-add something like that if we use it often.

### DIFF
--- a/ichnaea/alembic/versions/000000000000_base.py
+++ b/ichnaea/alembic/versions/000000000000_base.py
@@ -13,7 +13,6 @@ from alembic import op
 import sqlalchemy as sa
 from sqlalchemy import text
 
-from ichnaea.conf import settings
 
 DBCreds = namedtuple("DBCreds", "user pwd")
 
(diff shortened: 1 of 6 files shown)


ichnaea/alembic/versions/000000000000_base.py
```diff
@@ -29,60 +28,13 @@
...
-    grant_stmt = text("GRANT delete, insert, select, update ON *.* TO :user")
-    added = False
-    for cred in creds.values():
-        if cred.user not in userids:
-            conn.execute(create_stmt.bindparams(user=cred.user, pwd=cred.pwd))
-            conn.execute(grant_stmt.bindparams(user=cred.user))
-            userids.add(cred.user)
-            added = True
-    if added:
-        conn.execute("FLUSH PRIVILEGES")
```
willkg: I decided it's not great to have the migrations look at the environment variables and then add mysql db users. Seems like it's better for an ops person to explicitly add them for server environments. For local dev and testing environments, we just use the root account.

ichnaea/alembic/versions/000000000000_base.py
```diff
@@ -29,60 +28,13 @@
...
-    if not ("test",) in result:
-        stmt = text(
-            """\
-INSERT INTO api_key
-(valid_key, allow_fallback, allow_locate)
-VALUES
-('test', 0, 1)
-"""
-        )
-        conn.execute(stmt)
```
willkg: I removed this. We have a script for creating api keys now. That makes it easier for a developer to build the keys they need and script things.

ichnaea/scripts/apikey.py
```diff
@@ -0,0 +1,105 @@
...
+
+    db = configure_db("rw")
+    with db_worker_session(db) as session:
+        try:
+            session.execute(
+                insert(ApiKey.__table__).values(
+                    valid_key=key,
+                    allow_fallback=False,
+                    allow_locate=True,
+                    allow_region=True,
```
willkg: This hard-codes most of the values except the key. I wasn't sure whether to expose everything or not. I didn't want to create an annoying interface for it and wasn't sure what the predominant use cases would be. 

Maybe it's better to have subcommands for specific scenarios? Maybe "build an api key from this ini file" or something like that?

I figure we can defer that to when it comes up.

jwhitlock: I'm surprised the models don't have code-side or database-side defaults defined, which is really common in the Django world, but not in the SQLAlchemy world apparently. I'd prefer that the default values live on ``ApiKey`` eventually. It doesn't need to be in this PR.

I think we should use ops defaults for API keys (see private repo https://github.com/mozilla-services/cloudops-docs/tree/master/Services/Location#adding-new-api-key). That would add:

* ``maxreq=100000``
* ``store_sample_locate=100``
* ``store_sample_submit=100``

I think this would fulfill current needs, until we implement commercial user features like #890.

willkg: I don't understand sqlalchemy or how things are implemented. For example, there's an ApiKey and also a Key. One is db-backed and the other is cache-backed. Also, there is a factory for each model in the tests. So if you adjust one, you have to adjust it in four different places. That's curious. Why have two models for the same data? How do you know which one to use in the code? Why is there no abstraction layer for manipulating api key data? Why are default values strewn about? Anyhow... I figured I'd look into that at some later point.

Using ops defaults  is interesting, but I'm hesitant. This api key cli is currently just for the local dev environment. If we add ops defaults, I worry that suggests it should be used as a cli in production. I don't want to mix those two use cases because I suspect they have very different needs. That can get figured out in #846 .

Having said that, we should add the two store things because we need that for development. I'll add that now.

willkg:
Thank you!

PR #923, opened 2019-09-23
Update to Shapely 1.6.4.post2

jwhitlock:
* Shapely 1.5.17 → 1.6.4.post2: Optional features requiring ``numpy`` are gathered into the ``[vectorized]`` extra. Speedups based on the GEOS library are now enabled by default, so ``speedups.enable()`` is not needed. ``geos-config`` is only called at install, not at runtime. Add ``.split()`` and ``.polylabel()`` functions.

Shapely is used for the region lookup API. Tests pass, and speedups are enabled in the container:

```
$ docker-compose run --rm --entrypoint python app  -c "import shapely.speedups; print(shapely.speedups.enabled)"
True
```


ichnaea/__init__.py
```diff
@@ -1,12 +1,5 @@
...
 
-from shapely import speedups
-
 warnings.simplefilter("ignore", DeprecationWarning)
 warnings.simplefilter("ignore", PendingDeprecationWarning)
-
-# Enable shapely speedups
-if speedups.available:
-    speedups.enable()
-del speedups
```
willkg: Yay!

Issue #870, opened 2019-09-09
switch to a single mysql client library


willkg:
Ichnaea currently uses mysql-connector-python and PyMySQL as mysql client libraries. You can specify the db client library to use in settings.

The webapp uses gevent, so we still need a gevent-friendly mysql client. Otherwise, I think some of the reasons why Ichnaea does this have gone away over time.

This issue covers switching to a single mysql client library.

jwhitlock:
The [SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/dialects/mysql.html#module-sqlalchemy.dialects.mysql.mysqlconnector) say:

> The MySQL Connector/Python DBAPI has had many issues since its release, some of which may remain unresolved, and the mysqlconnector dialect is not tested as part of SQLAlchemy’s continuous integration. The recommended MySQL dialects are mysqlclient and PyMySQL.

Django [recommends](https://docs.djangoproject.com/en/2.2/ref/databases/#mysql-db-api-drivers) [mysqlclient](https://github.com/PyMySQL/mysqlclient-python).

willkg:
@jwhitlock and I had a long soul-searching conversation that I will summarize.

Ichnaea currently uses both PyMySQL and mysqlconnector mysql client libraries. They're used in different places, but both have to be installed. The 1.5 branch which is currently in production uses only PyMySQL, so this is unique to the master branch which hasn't been deployed anywhere.

Ichnaea's webapp runs using gunicorn and the gevent worker, so whatever client we end up with needs to support gevent. Generally, pure Python libraries support gevent.

It'd be a  project to change Ichnaea to not use gevent, so that isn't on the table right now.

@jwhitlock wondered whether we could use mysqlclient-python which Django recommends, but that doesn't support gevent, so that's out as an option.

The db code encodes this notion of a "sync" transport. When something configures a db using "sync", it gets the mysqlconnector client. The sync transport is used by `ichnaea/scripts/datamap.py` which runs via cron on the Celery scheduler node (covered in #840). It's also used by `ichnaea/scripts/dump.py` which we think is a script admins run to dump some db contents for importing into a local dev environment. We'll probably remove that in #503. It's also used in `ichnaea/conftest.py` but probably only to test those two scripts. Thus, we think we can remove the "sync" transport code.

Given all that, we're going to switch to the PyMySQL client for everything and remove code that switches between multiple clients.

We think that'll probably require the following work:

1. remove the mysqlconnector client as a requirement
[...]

PR #926, opened 2019-09-24
Update to PyMySQL 0.9.3

### LINKED ISSUE #863: Update PyMySQL
Update to the latest PyMySQL. There are several test failures when updating, so code or configuration changes may be needed.

jwhitlock:
* PyMySQL 0.7.11 → 0.9.3: Change defaults for compatibility with [mysqlclient](https://github.com/PyMySQL/mysqlclient-python), improve support for encrypted passwords. See [changelog](https://github.com/PyMySQL/PyMySQL/blob/master/CHANGELOG).

The smallest change to support this version is to set connection arguments that return to PyMySQL 7.0's default behavior.

Fixes #863. The larger changes to support just one MySQL client are in #870.


ichnaea/db.py
```diff
@@ -156,6 +157,12 @@ def __init__(self, uri, pool=True, transport="default"):
             options["connect_args"]["ssl_version"] = PROTOCOL_TLSv1
             # Needed for SSL
             options["connect_args"]["ssl_ca"] = certifi.where()
+        elif DB_TRANSPORTS[transport] == 'pymysql':
+            # PyMySQL 8.0 changed for compatibility with mysqlclient
+            # Use 7.0's binary prefixes, for selecting and inserting binary data
```
willkg: I think you mean `0.8.0` and `0.7.0` here.

jwhitlock: yep 🤦‍♂

jwhitlock:
Force pushed for formatting and to get 0.8 / 0.7 right in code and commit message. Tests finally pass, and were failing hard before the changes to ``db.py``.

PR #927, opened 2019-09-24
Update to mysql-connector-python 8.0.17

### LINKED ISSUE #881: Update mysql-connector-python and protobuf
Update mysql-connector-python, the MySQL library used for some connections.

See #870, which may eliminate this driver.

jwhitlock:
* mysql-connector-python 8.0.16 → 8.0.17: Fix bug with MySQL 5.7, Python 3, and the C extension (``use_pure=False``). The C extension fails with our code, so ichnaea continues to set ``use_pure=True`` to use the Python implementation. See [8.0.17 release notes](https://dev.mysql.com/doc/relnotes/connector-python/en/news-8-0-17.html) for more.

This version adds a new deprecation warning:
```
  /usr/local/lib/python3.6/site-packages/mysql/connector/connection.py:126: PendingDeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5
    os_ver = "-".join(platform.linux_distribution()[0:2]) # pylint: disable=W1505
```

Updated dependencies:
* protobuf 3.7.1 → 3.9.2: Latest version. Python code is generated, no release notes, just [docs](https://developers.google.com/protocol-buffers/docs/reference/python-generated).

Fixes #881. Switching to a single MySQL client is in #870. I'm not a fan of either of the two we are using.


jwhitlock:
After some discussion, we're going to work on removing this driver (#870) rather than updating and confirming it still works.

Issue #913, opened 2019-09-18
fix celery configuration


willkg:
There's some configuration being done in `ichnaea/taskapp/settings.py`. In prod, the script sets the `worker_concurrency` to double the number of cores. Seems like that got lost at some point.

This issue covers adding that as an environment variable.

While doing that, we should look at what other things should be configurable.

willkg:
https://docs.celeryproject.org/en/latest/userguide/configuration.html

Also, we probably want to do this after upgrading celery in issue #862 .

jwhitlock:
With celery 4, the logging level is now set on the command line, like ``--loglevel=info``. I find ``info`` useful for ensuring things are working in development, but ``warning`` might be right for production.  Another setting?

Issue #888, opened 2019-09-10
switch everything to use json module


willkg:
Ichnaea uses json in some places and simplejson in others.

This issue covers figuring out whether there's a compelling reason to keep doing that. If there is, we should document why a developer should use simplejson instead of json so there's convention and consistency. If there isn't a reason, then we should switch everything to use one or the other.

jwhitlock:
I support switching ichnaea code to use the standard library ``json``.

From the [simplejson docs](https://simplejson.readthedocs.io/en/latest/):

> `simplejson` exposes an API familiar to users of the standard library `marshal` and `pickle` modules. It is the externally maintained version of the json library contained in Python 2.6, but maintains compatibility with Python 2.5 and (currently) has significant performance advantages, even without using the optional C extension for speedups. `simplejson` is also supported on Python 3.3+.

There's a lot of benchmarks from 2011-2014 showing speed improvements in ``simplejson`` version ``json`` in Python 2.6 and 2.7. All appearances are that these speed improvements were merged into the standard library ``json``. It might be worth a blog article, if you are inclined.

``datadog`` requires ``simplejson`` until 0.24.0 (update issue #867).

jwhitlock:
I started to look into this a little more. Most uses of ``simplejson`` would be easy enough to change. In Python 2, ``json.dumps`` returned a bytestring (type ``str``), and in Python 3, it returns a unicode string.  ``simplejson`` returns a unicode string in both cases, so the code is careful to change to and from utf-8 encoded bytestrings as it passes data to and from Redis.

A slightly harder part is a Celery serialization method ``"internal_json"`` that uses simplejson's functions to handle content encoded in UTF-8:

https://github.com/mozilla/ichnaea/blob/a6a5426413219878e8d32152142beb2a1e2cb89e/ichnaea/taskapp/config.py#L28-L34

This is used in the task app as the result and task serializer:

https://github.com/mozilla/ichnaea/blob/a6a5426413219878e8d32152142beb2a1e2cb89e/ichnaea/taskapp/settings.py#L79-L81

In Celery 3, the default serializer was ``pickle``, and [Celery 4 switchs to ``json``](https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html#json-is-now-the-default-serializer). We should be able to switch from ``internal_json`` to the built-in ``json``, as part of #861 (lastest celery 3) and then #862 (celery 4).

Issue #862, opened 2019-09-06
Update to Celery 4


jwhitlock:
Update to Celery 4, including the code changes needed for Celery 4. See the guide for [Upgrading from Celery 3.1](https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html#upgrading-from-celery-3-1).

Tasks:
* [x] Convert to [lowercase setting names](https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html#v400-upgrade-settings)
* [x] Remove ``simplejson``-based serializer (see #888)
* [x] Read rest of upgrade notes, fix isses
* [x] Update to Celery 4.0.x, redis 3.x (see #874)
* [ ] Update to latest Celery, redis

There are related celery issues, which can be done in parallel or after this is complete:

* Use mozlog format in Celery logs (#909)
* Fix celery configuration (#913)

PR #928, opened 2019-09-26
Update to celery 4.0.2

jwhitlock:
Requirement updates:
* celery 3.1.26.post2 → celery 4.0.2: Many breaking changes. See [what's new in 4.0](https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html).

Compatible dependency upgrades:
* amqp 1.4.9 → 2.5.1: Drop Python 2.6 and 3.4, other fixes. See [changelog](https://github.com/celery/py-amqp/blob/master/Changelog).
* billiard 3.3.0.23 → 3.5.0.5: Drop Python 2.6, other fixes. See [changelog](https://github.com/celery/billiard/blob/master/CHANGES.txt).
* kombu 3.0.37 → 4.1.0: Requires amqp 2.x, drops anyjson, drops unsupported transports like Django ORM, add new transports like SQS. JSON serializer calls ``__json__`` method if available. See [changelog](https://docs.celeryproject.org/projects/kombu/en/stable/changelog.html)
* vine 1.3.0: New requirement of amqp 2.1.0. See https://github.com/celery/vine.

Convert code to Celery 4:
* Use lowercase settings names
* Use ``--loglevel=INFO`` to see tasks as they fly by
* Drop ``BaseTask`` attributes that are the default or are loaded from the configuration
* Adopt other ``BBB`` notes

This is progress on #862, but there is more work to finish the celery update. My next step is getting rid of ``simplejson`` (#888), and the work on configuration (#913) can be done in parallel.


ichnaea/taskapp/task.py
```diff
@@ -13,12 +13,6 @@
 class BaseTask(Task):
     """A base task giving access to various outside connections."""
 
-    # BBB: Celery 4 non-underscore attributes
-    abstract = True
-    acks_late = False
-    ignore_result = True
-    max_retries = 3
```
willkg: What happened to these? Are they accounted for somewhere else?

jwhitlock: Yep, these are no longer needed, set in configuration, or already the default.

[The Task base class no longer automatically registers tasks](https://docs.celeryproject.org/en/latest/history/whatsnew-4.0.html#the-task-base-class-no-longer-automatically-register-tasks). Instead, they can be registered manually or with the ``@celery_app.task`` decorator. This change also means that the ``abstract`` attribute of the task no longer has any effect.

The default for [tasks_acks_late](https://docs.celeryproject.org/en/latest/userguide/configuration.html#task-acks-late) is Disabled, so this ``acks_late = False`` is redundant.

The default for [task_ignore_result](https://docs.celeryproject.org/en/latest/userguide/configuration.html#task-ignore-result) is Disabled, but we have ``task_ignore_result = True`` in ``ichnaea.taskapp.settings``, so it is redundant here.

The default for [max_retries](https://github.com/celery/celery/blob/8cf8fae70630954cff3448485588bbe2a77ff3ab/celery/app/task.py#L178-L180) is ``3``, but it is undocumented and you have to peek at the code.

Issue #840, opened 2019-08-14
run map tile generation in server environment


willkg:
Previously, map tile generation (`ichnaea/scripts/datamap.py`) would get run like this via cron on the scheduler node:

```
17 0 * * * location ICHNAEA_CFG=/etc/location.ini /data/ichnaea/bin/location_map --create --concurrency=<%= @processorcount.to_i %> --upload --output=<%= @tiles_directory %> > /var/log/location/cronjobs/map.log 2>&1
```

We don't currently have cron running in the docker container. We should probably move this to a celery task or some other scheduler.

hannosch:
This ran on the celery scheduler, as that machine was otherwise underutilized. The datamaps stuff benefits greatly from running parts of it in parallel processes and probably runs a couple hours. It dumps a large result from MySQL locally on disk, post-processes it, crates half a million or so tiny png files and than uploads the new and changed files to an Amazon S3 bucket. That bucket is exposed via cloudfront as a website and gets used a tile layer in the Mapbox code.

The script takes quite long, needs the large SQL dump and those files all on the same disk and does its own multiprocessing. That made it a bad fit for the normal Celery or any other task queue.

willkg:
This should get figured out, but we think it'll be fine if we push it off until after the new infrastructure is up and running.

willkg:
@ckolos said he can run this trivially.

However, I changed how ichnaea works in the Docker image and it no longer gets installed as a Python library, thus the `location_map` script isn't "built".

We need to fix how that gets run in the ichnaea Docker image and then we can hand it off to @ckolos and he can operationalyze it.

jwhitlock:
I've checked that the script runs in the dev environment from a Docker instance, from a copy of stage's datamap tables. It took 70-80 minutes to generate tiles, and triggered my machine to reboot at least once. I think production is at least 10x larger, but I doubt it will take 10x longer.

I also tried uploading to a development S3 bucket. It took 80 minutes to generate tiles, then failed due to lack of credentials. I'm adding some logic to check that the caller can write to the bucket before generating the tiles. The script is now running with upload (no concurrency), I'll update with how long it took.

My next step are to add more logging, tune metrics, and document the data flow (issue #1404). Then, I'll run it manually in stage, from the admin or cron server. Once we have the process figured out, we can operationalize it to run periodically.

jwhitlock:
I've merged changes and deployed to staging. 

[...]

PR #931, opened 2019-09-27
Remove mysqlconnector requirement and usage

### LINKED ISSUE #870: switch to a single mysql client library

willkg:
mysqlconnector was added a couple of years ago for the datamap and dump
scripts. However, it adds a lot of complexity to the codebase and forces
us to keep tabs on two different mysql libraries.

A lot of things have changed in the last couple of years in the world of
Python and mysql clients and we think the compelling reasons to have two
are gone now. Thus we're going to simplify the situation and switch back
to PyMySQL for all mysql connections.

Fixes #870

To test:

1. `make build` and `make setup`
2. `make lint` and `make test`
3. `make run` and run through the site--though it's all static pages for the most part
4. `make runcelery` and make sure the scheduler and workers do things

This doesn't test the datamap and dump scripts outside of whatever gets tested in the test suite. We'll test datamap more thoroughly as part of #503 and #840. If there are issues, we can figure them out there.


willkg:
Thank you!

PR #933, opened 2019-10-02
centralize database uri for alembic migrations

### LINKED ISSUE #904: nix ddl uri
There's a `DB_DDL_URI` configuration variable which sets up the uri for running a few things:

1. setting up the db for tests
2. `create_db` and `drop_db` which are only used in the local dev environment
3. alembic migrations

Two of those three things are development-related. If we're planning to continue running migrations manually, then we can probably nix the ddl uri. The other things can all use the readwrite uri.

This issue covers figuring this out.

willkg:
There are a few ways we run alembic migrations:

1. running the alembic command from the shell to run migrations
   in a local dev environment
2. running the alembic command from the shell to run migrations
   in a server environment
3. running alembic subcommands from ichnaea/scripts/db.py
4. running alembic subcommands for test setup

We want all of these cases to pull the database uri from an environment
variable. This changes the code such that all of these use cases use the
`SQLALCHEMY_URL` environment variable.

In the local dev environment, this value is in `docker/config/app.env`. In a server environment, they'll be provided on the admin node in the environment.

To test:

1. run `make build` and `make setup`
2. run `make test`
3. run `make shell` and then at the prompt `alembic current -v`

All of those situations should work.

Fix #904.


willkg:
One thing I wasn't sure about was how to document `SQLALCHEMY_URL`. It's not a setting used by the app--it's only used by alembic. It doesn't make sense to me to require it in the app config. We could add a `DatabaseConfig` class in `ichnaea/db.py` and autodoc that. I think the only case where this comes up is when admin are running `alembic` in a server environment. Thoughts?

docker/config/app.env
```diff
@@ -10,4 +10,6 @@ REDIS_URI=redis://redis:6379/0
 # DB
 DB_READONLY_URI=mysql+pymysql://root:location@mysql:3306/location
 DB_READWRITE_URI=mysql+pymysql://root:location@mysql:3306/location
-DB_DDL_URI=mysql+pymysql://root:location@mysql:3306/location
+
+# DB url for running alembic migrations
+SQLALCHEMY_URL=mysql+pymysql://root:location@mysql:3306/location
```
willkg: I named this closer to what alembic calls it figuring maybe that makes it clearer that it's an alembic thing and not an app configuration thing.

It is in `app.env`. I'm kind of loathe to create another env file, though. I'm not sure how to make this clearer without making it more complex. Maybe we rename `app.env` to `local_dev.env`?

ichnaea/conftest.py
```diff
@@ -90,29 +90,30 @@ def package():
...
+            state = dict(key.__dict__)
+            del state["_sa_instance_state"]
+            conn.execute(key.__table__.insert().values(state))
             trans.commit()
 
 
+def setup_tables(engine):
+    # Stamp the latest alembic version
+    alembic_main(["stamp", "base"])
+    alembic_main(["upgrade", "head"])
```
willkg: Calling these like this will run alembic as if it was being run from the command line, so it picks up `alembic.ini` and `env.py` and all the bits.

ichnaea/conftest.py
```diff
@@ -134,7 +135,7 @@ def setup_database():
         pass
 
     # Clean up the tables and set them up
-    db = configure_db("ddl")
+    db = configure_db("rw")
```
willkg: I should change this to use `get_sqlalchemy_url()` and not the `rw` db thing.

ichnaea/db.py
```diff
@@ -17,36 +16,47 @@
 from ichnaea.conf import settings
 
 
-DB_TYPE = {
-    "ddl": settings("db_ddl_uri"),
-    "ro": settings("db_readonly_uri"),
-    "rw": settings("db_readwrite_uri"),
-}
+DB_TYPE = {"ro": settings("db_readonly_uri"), "rw": settings("db_readwrite_uri")}
```
willkg: black did this. I find it unhelpful, but not enough to fight it.

ichnaea/db.py
```diff
@@ -17,36 +16,47 @@
...
+class SqlAlchemyUrlNotSpecified(Exception):
+    """Raised when SQLALCHEMY_URL is not specified in environment."""
 
-    script_location = os.path.join(os.path.dirname(__file__), "alembic")
-    assert os.path.exists(script_location)
-    cfg.set_section_option("alembic", "script_location", script_location)
-    cfg.set_section_option("alembic", "sqlalchemy.url", settings("db_ddl_uri"))
-    return cfg
+    def __init__(self, *args, **kwargs):
+        super().__init__("SQLALCHEMY_URL is not specified in the environment")
```
willkg: The first argument is `args[0]` which gets used as the msg. I figure this lets me (ab)use the exception type, but have a helpful message regardless of where it's raised.

ichnaea/db.py
```diff
@@ -17,36 +16,47 @@
...
+def configure_db(type_=None, uri=None, _db=None, pool=True):
     """
     Configure and return a :class:`~ichnaea.db.Database` instance.
 
+    :param type_: one of "ro" or "rw"
+    :param uri: the uri to connect to; if not provided, uses ``type_``
+        parameter
     :param _db: Test-only hook to provide a pre-configured db.
+    :param pool: True for connection pool, False for no connection pool
+
```
willkg: At this point, this is kind of a mess of a function. I thought about trying to make it less messy, but didn't have any good ideas that seemed worthwhile to spend time on. I figure we can push this off to later.

ichnaea/db.py
```diff
@@ -165,17 +175,18 @@ def ping_session(db_session):
...
+        environment variable
 
     :raises sqlalchemy.exc.ProgrammingError: if database already exists
 
     Note: This is mysql-specific.
 
     """
     if uri is None:
-        uri = DB_TYPE["ddl"]
+        uri = os.environ.get("SQLALCHEMY_URL", "")
```
willkg: This should use `get_sqlalchemy_url()`.

ichnaea/db.py
```diff
@@ -207,15 +207,16 @@ def drop_db(uri=None):
...
+        environment variable
 
     :raises sqlalchemy.exc.InternalError: if database does not exist
 
     Note: This is mysql-specific.
 
     """
     if uri is None:
-        uri = DB_TYPE["ddl"]
+        uri = os.environ.get("SQLALCHEMY_URL", "")
```
willkg: This should use `get_sqlalchemy_url()`.

ichnaea/scripts/datamap.py
```diff
@@ -59,7 +59,7 @@ def export_file(filename, tablename, _db=None, _session=None):
         )
     )
 
-    db = configure_db("ro", _db=_db)
+    db = configure_db("ro", _db=_db, pool=False)
```
willkg: Now that we have a `pool` argument, we can pass it in here and make sure it's not using a pool.

ichnaea/scripts/dump.py
```diff
@@ -110,7 +110,7 @@ def main(argv, _db=None, _dump_file=dump_file):
 
     configure_logging()
 
-    db = configure_db("ro", _db=_db)
+    db = configure_db("ro", _db=_db, pool=False)
```
willkg: Now that we have a `pool` argument, we can pass it in here and make sure it's not using a pool.

docker/config/app.env
willkg: I'm going to switch it. It represents an environment (e.g. local dev environment, stage, prod)--not a service (e.g. webapp, celery).

willkg:
^^^ I fixed the issues I found in self-review. This is good to review now.

willkg:
The changes herein affect the local dev environment and tests and both of those look ok. If there are issues with server environments, we'll discover and fix those in future PRs.

self-r+

PR #935, opened 2019-10-02
Remove IRC references

willkg:
irc.mozilla.org is going away soon. This removes references to it.


willkg:
self-r+ since this only affects docs.

PR #936, opened 2019-10-02
Add CELERY_WORKER_CONCURRENCY setting and fix Celery logging

### LINKED ISSUE #909: fix celery logging
Celery logging in the local dev environment doesn't seem to be telling me much. I think we want it to log at DEBUG level for ichnaea things so we can more easily see what's going on.

Celery logging in the server environment needs to use the mozlog format.

Issue #862 covers upgrading to Celery 4. I think we should do that first, then rework logging.

### LINKED ISSUE #913: fix celery configuration

willkg:
This adds a `CELERY_WORKER_CONCURRENCY` setting.

This redoes Celery logging configuration so it uses the same configuration as the rest of Ichnaea including whether or not to use mozlog.

Fixes #909.
Fixes #913.

To test:

1. Run `make runcelery`.

   Should spin up one celery worker with concurrency: 1. You can tell by looking at the output when it spits out a `[config]` section to the right of a giant ASCII C.

   Logging output should look "normal" for the scheduler and worker. Like this:
   ```
   worker_1     | 2019-10-02 15:23:27,433 INFO  [celery.app.trace] - Task ichnaea.data.tasks.update_incoming[7a798f9e-7b81-4f56-b537-9979cfa60cd6] succeeded in 0.061439532990334556s: None
   ```
2. Add `CELERY_WORKER_CONCURRENCY=2` and `LOCAL_DEV_ENV=False` to `my.env` and then run `make runcelery`.

   Should spin up one celery worker with concurrency: 2.

   Output is in mozlog format after the scheduler and worker have started up. Like this:
   ```
   worker_1     | {"Timestamp": 1570029523458684160, "Type": "celery.worker.consumer.connection", "Logger": "ichnaea", "Hostname": "3fa9d5c12283", "EnvVersion": "2.0", "Severity": 6, "Pid": 7, "Fields": {"msg": "Connected to redis://redis:6379/0"}}
   ```

### DIFF
--- a/ichnaea/log.py
+++ b/ichnaea/log.py
@@ -22,9 +22,11 @@ def configure_logging():
 
     if local_dev_env:
         level = "DEBUG"
+        celery_level = "INFO"
         handlers = ["console"]
     else:
         level = "INFO"
+        celery_level = "WARNING"
         handlers = ["mozlog"]
 
     logging_config = {
@@ -51,6 +53,7 @@ def configure_logging():
         },
         "loggers": {
             "alembic": {"propagate": False, "handlers": handlers, "level": level},
+            "celery": {"propagate": False, "handlers": handlers, "level": celery_level},
             "ichnaea": {"propagate": False, "handlers": handlers, "level": level},
         },
         "root": {"handlers": handlers, "level": "WARNING"},
--- a/ichnaea/taskapp/app.py
+++ b/ichnaea/taskapp/app.py
@@ -3,14 +3,29 @@ Holds global celery application state and startup / shutdown handlers.
 """
 from celery import Celery
 from celery.app import app_or_default
-from celery.signals import beat_init, worker_process_init, worker_process_shutdown
+from celery.signals import (
+    beat_init,
+    worker_process_init,
+    worker_process_shutdown,
+    setup_logging,
+)
 
+from ichnaea.log import configure_logging
 from ichnaea.taskapp.config import (
     configure_celery,
     init_beat,
     init_worker,
     shutdown_worker,
 )
+
+
+@setup_logging.connect
+def setup_logging_process(loglevel, logfile, format, colorize, **kwargs):
+    """Called at scheduler and worker setup.
+
+    Configures logging using the same configuration as the webapp.
+    """
+    configure_logging()
 
 
 @beat_init.connect
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -13,3 +13,6 @@ DB_READWRITE_URI=mysql+pymysql://root:location@mysql:3306/location
 
 # DB url for running alembic migrations
 SQLALCHEMY_URL=mysql+pymysql://root:location@mysql:3306/location
+
+# Limit Celery workers to 1 in the local dev environment
+CELERY_WORKER_CONCURRENCY=1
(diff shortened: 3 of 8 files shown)


ichnaea/log.py
```diff
@@ -22,9 +22,11 @@ def configure_logging():
 
     if local_dev_env:
         level = "DEBUG"
+        celery_level = "INFO"
         handlers = ["console"]
     else:
         level = "INFO"
+        celery_level = "WARNING"
```
willkg: @jwhitlock suggested setting this to WARNING in stage/prod and INFO in local dev environment. I figured I'd follow that advice. If we ever need to change it, we can make it a setting or something.

ichnaea/taskapp/app.py
```diff
@@ -13,6 +19,15 @@
 )
 
 
+@setup_logging.connect
+def setup_logging_process(loglevel, logfile, format, colorize, **kwargs):
+    """Called at scheduler and worker setup.
+
+    Configures logging using the same configuration as the webapp.
+    """
+    configure_logging()
```
willkg: This will override Celery's normal logging setup which is what we want.

ichnaea/log.py
jwhitlock: When testing the prod-like settings, the scheduler still says it is logging as ``INFO``:

```
scheduler_1  | __    -    ... __   -        _
scheduler_1  | LocalTime -> 2019-10-02 16:06:15
scheduler_1  | Configuration ->
scheduler_1  |     . broker -> redis://redis:6379/0
scheduler_1  |     . loader -> celery.loaders.app.AppLoader
scheduler_1  |     . scheduler -> celery.beat.PersistentScheduler
scheduler_1  |     . db -> /var/run/location/celerybeat-schedule
scheduler_1  |     . logfile -> [stderr]@%INFO
scheduler_1  |     . maxinterval -> 5.00 minutes (300s)
```

I believe this is because of ``--loglevel INFO`` in ``scripts/run_scheduler.sh``. However, the output is consistent with ``WARNING`` level output. This is confusing, may bite us later.

I overrode ``celery_level = "INFO"`` in the prod-like mode, and I get the many expected lines of the ugly mozlog format, so this ``ichnaea/log.py`` code is setting the log level, not the command line.

ichnaea/taskapp/app.py
```diff
@@ -13,6 +19,15 @@
 )
 
 
+@setup_logging.connect
+def setup_logging_process(loglevel, logfile, format, colorize, **kwargs):
```
jwhitlock: I would guess that ``loglevel`` is the ``--loglevel`` passed from the command line. If that's the case, we could configure the log level from ``scripts/run_celery.sh`` and ``scripts/run_scheduler.sh``. I'd support waiting until will configure from an environment variable to take that step.

willkg: I forgot we set it with a flag there. I'll nix the `--loglevel` argument in the `run_*` scripts.

willkg:
Thank you!

PR #939, opened 2019-10-03
Update to Python 3.7

### LINKED ISSUE #934: update to Python 3.7
Ichnaea is currently running on Python 3.6. We should update it to Python 3.7.

willkg:
This updates the image we're using to Python 3.7.4. That's the current latest version available:

https://hub.docker.com/_/python?tab=tags&page=1&name=3.7

To test:

1. `make test`
2. `make lintfix`
3. `make lint`
4. `make run` and `make runcelery`

Everything seems fine. I think we might have hit issues before we updated all the requirements and renamed the "async" module. But now that we've done that, this is pretty straight-forward.

Fixes #934

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.9-slim
+FROM python:3.7.4-slim
 
 # Set up user and group.
 ARG groupid=10001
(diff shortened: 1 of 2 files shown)


willkg:
Thank you!

Issue #830, opened 2019-08-07
switch to markus for metrics


willkg:
Ichnaea sends metrics to Datadog using the datadog Python library. We're probably going to switch off of Datadog, so we'd need to rework metrics to use the statsd library.

There's some code in `ichnaea/log.py` to create a debug datadog client.

We should look at switching to using [Markus](https://markus.readthedocs.io/) which provides a convenient abstraction layer over metrics generation. That'll make it easier to switch from datadog to statsd in the future--it'll just be a configuration change.

hannosch:
The metrics code relies heavily on tags for each metric. IIRC those are a datadog specific addition to statsd, so that might be an issue.

willkg:
My bad--I misunderstood what we were doing. We're not switching to the statsd protocol--we're sticking with the datadog enhancements to statsd. So tags and all that should work fine going forward. Yay!

Still, I think switching to Markus has some value. It helped us a lot in development and testing with Socorro, Antenna, Buildhub2, and Tecken. It's got some niceties.

PR #941, opened 2019-10-03
Add STATSD_PORT to configuration

### LINKED ISSUE #940: make statsd port configurable
There is a `STATSD_HOST` configuration setting, but the port is hard coded:

https://github.com/mozilla/ichnaea/blob/a43e16a40216ab6ccfedd1d9f416e98399021bba/ichnaea/log.py#L106

We should make the port configurable, too.

willkg:
Fixes #940


willkg:
I'm not entirely sure how to test this since there's no output indicating what happened in the logs. I was thinking I'd get this in now and then when we rework statsd things with markus in #830 we could make it clearer as to what's going on then.

willkg:
Thank you!

PR #942, opened 2019-10-04
make mysql: connect to the location database

jwhitlock:
When running ``make mysql``, connect to the ``location`` database, so that a ``connect location`` is not needed for querying the regular database.


jwhitlock:
Thanks!

PR #943, opened 2019-10-04
update_statregion: Use ORM, not bulk_*_mappings

jwhitlock:
MySQL returns the number of rows affected by an ``UPDATE``, and ``bulk_update_mappings`` raises a ``StaleDataError`` if the affected rows are less than the rows sent.  This means that the update fails if the radio count is the same for any region. Instead, return to using the ORM methods, which sends several ``UPDATE``s, but only when the data changes.

Without this change, the new test ``test_update_no_change`` fails when ``update_statregion`` is called with no changes to the ``CellArea`` counts:

```
...
  File "/app/ichnaea/data/stats.py", line 156, in _update_stats
    session.bulk_update_mappings(RegionStat, updates)
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/orm/session.py", line 2849, in bulk_update_mappings
    mapper, mappings, True, False, False, False, False
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/orm/session.py", line 2888, in _bulk_save_mappings
    transaction.rollback(_capture_exception=True)
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 153, in reraise
    raise value
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/orm/session.py", line 2873, in _bulk_save_mappings
    update_changed_only,
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/orm/persistence.py", line 180, in _bulk_update
    bookkeeping=False,
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/orm/persistence.py", line 1029, in _emit_update_statements
    % (table.description, len(records), rows)
sqlalchemy.orm.exc.StaleDataError: UPDATE statement on table 'region_stat' expected to update 4 row(s); 0 were matched.
```

This also issues a ``ROLLBACK``, so it is not enough to catch the error.  Any rows with the same data would need to be removed from the update list, which would require querying and comparing to the current values, which is exactly what using the SQLAlchemy ORM does. This optimization to use bulk inserts and updates doesn't seem worth it for a table of 200-300 records that is populated in a task.

### DIFF
--- a/ichnaea/data/tests/test_stats.py
+++ b/ichnaea/data/tests/test_stats.py
@@ -1,5 +1,7 @@
 from datetime import timedelta
 
+import pytest
+
 from ichnaea.cache import redis_pipeline
 from ichnaea.data.tasks import cleanup_stat, update_statcounter, update_statregion
 from ichnaea.models import Radio, RegionStat, Stat, StatCounter, StatKey
@@ -131,13 +133,25 @@ class TestStatCleaner(object):
 
 class TestStatRegion(object):
     def test_empty(self, celery, session):
+        """update_statregion exits early with no data."""
         update_statregion.delay().get()
         stats = session.query(RegionStat).all()
         assert stats == []
 
-    def test_update(self, celery, session):
+    def test_null_region_no_stat(self, celery, session):
+        """update_statregion ignores stations with no region."""
         area = CellAreaFactory(radio=Radio.gsm, num_cells=1)
         area.region = None
+        wifi = WifiShardFactory()
+        wifi.region = None
+        session.flush()
+
+        update_statregion.delay().get()
+        stats = session.query(RegionStat).all()
+        assert stats == []
+
+    def test_insert(self, celery, session):
+        """update_statregion creates RegionStats for new regions."""
         BlueShardFactory.create_batch(2, region="CA")
         BlueShardFactory.create_batch(3, region="GB")
         CellAreaFactory(radio=Radio.gsm, region="DE", num_cells=1)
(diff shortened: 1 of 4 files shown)


ichnaea/data/tests/test_stats.py
```diff
@@ -167,3 +167,10 @@ def test_update(self, celery, session):
                 assert values == (0, 0, 0, 3, 0)
             elif stat.region == "US":
                 assert values == (0, 0, 0, 0, 6)
+
+    def test_update_no_change(self, celery, session):
+        """update_statregion is successful when there are no area changes."""
+        self.test_update(celery, session)
+        update_statregion.delay().get()
+        stats = session.query(RegionStat).all()
+        assert len(stats) == 4
```
willkg: Instead of calling `test_update` which then ties this test to that one, can we just do a minimal version of `test_update` in this test? Maybe something like this?:

```
   def test_update_no_change(self, celery, session):
        """update_statregion is successful when there are no area changes."""
        area = CellAreaFactory(radio=Radio.gsm, num_cells=1)
        area.region = None
        BlueShardFactory.create_batch(2, region="CA")
        BlueShardFactory.create_batch(3, region="GB")
        CellAreaFactory(radio=Radio.gsm, region="DE", num_cells=1)
        CellAreaFactory(radio=Radio.gsm, region="DE", num_cells=2)
        CellAreaFactory(radio=Radio.gsm, region="CA", num_cells=2)
        CellAreaFactory(radio=Radio.wcdma, region="DE", num_cells=3)
        CellAreaFactory(radio=Radio.lte, region="CA", num_cells=4)
        WifiShardFactory.create_batch(5, region="DE")
        WifiShardFactory.create_batch(6, region="US")
        wifi = WifiShardFactory()
        wifi.region = None
        session.add(RegionStat(region="US", blue=1, wifi=2))
        session.add(RegionStat(region="TW", wifi=1))
        session.flush()

        update_statregion.delay().get()
        stats = session.query(RegionStat).all()
        assert len(stats) == 4

        update_statregion.delay().get()
        stats = session.query(RegionStat).all()
        assert len(stats) == 4
```

jwhitlock: I like independent tests, good idea.

This new test covers that unchanged cell areas don't break ``update_statregion``. This could also be done by setting some ``RegionStat`` rows and calling ``update_statregion``. The existing test covers inserting new ``RegionStat`` rows.

 A few more things to test:

* An updated ``CellArea`` results in an updated ``RegionStat``
* If all the ``CellArea``s for a region are removed, then the ``RegionStat`` is deleted.

I may need to resurrect code coverage to guide these changes.

jwhitlock: Code coverage is interesting.

For the old code, using ``bulk_update_mappings``, code coverage wouldn't have highlighted any problems. It just shows branch misses for the case with no inserts, updates, and nothing to delete.

For the new code, the existing test covers everything but the case where there are no obsolete regions to delete. Running ``update_statregion`` twice covers this branch. In the original test, the ``TW`` region is an obsolete region that will be deleted.

The existing test includes a change to the ``US`` region. I could add more data that includes a region that does not update, or I could add multiple tests for each case - new region, updated region, same region, deleted region.  I have a preference for single purpose test cases.

jwhitlock:
The application code is the same, but the tests have been expanded and split up, and now there is a ``make testcoverage`` command to see the missing lines and branches.

jwhitlock:
Thanks for the double review, @willkg!

PR #944, opened 2019-10-08
Switch to Markus for metrics

willkg:
This removes the statsd-related scaffolding in favor of the Markus library. Markus makes it easier to configure a project for local development differently than server environments. It also comes with testing tools for asserting things about generated metrics.

This removes the statsd-related scaffolding and replaces it with equivalent Markus things.

The tests verify metrics pretty extensively, so to test this, I'd just run the tests.

Fixes #835


ichnaea/api/locate/tests/base.py
```diff
@@ -455,104 +469,124 @@ def test_api_key_blocked(self, app, data_queues, session):
...
+                "key:test",
+                "region:none",
+                "geoip:false",
+                "blue:many",
+                "cell:none",
+                "wifi:none",
             ],
-            timer=[("request", [self.metric_path, "method:post"])],
+        )
+        metricsmock.print_records()
```
willkg: Oops--this and all the other instances of this should get removed.

ichnaea/api/locate/tests/base.py
```diff
@@ -561,67 +595,85 @@ def test_cell_invalid_lac(self, app, data_queues):
...
-                (self.metric_type + ".request", [self.metric_path, "key:test"]),
-                ("request", [self.metric_path, "method:post", "status:200"]),
-            ]
+        assert metricsmock.has_record(
+            "incr",
+            self.metric_type + ".request",
+            value=1,
+            tags=[self.metric_path, "key:test"],
+        )
+        assert metricsmock.has_record(
```
willkg: I'm tempted to tweak markus to have `.has_incr`, `.has_gauge`, `.has_timing`, and `.has_histogram`. It reduces a lot of these by at least a line without losing any clarity.

@jwhitlock What do you think about that?

ichnaea/api/locate/tests/base.py
```diff
@@ -701,96 +764,118 @@ def test_cellarea_with_different_fallback(self, app, session, stats):
...
+        metricsmock.print_records()
+        assert metricsmock.has_record(
+            "incr",
+            self.metric_type + ".result",
+            value=1,
+            tags=[
+                "key:test",
+                "region:none",
+                "accuracy:high",
+                "fallback_allowed:false",
```
willkg: This one is curious. The original check argument is this:

```
(
    self.metric_type + ".result",
    "fallback_allowed:false",
    ["key:test", "region:none", "accuracy:high", "status:miss"],
),
```

The check code is here:

https://github.com/mozilla/ichnaea/blob/ba487a97bca4b6343ae0365267ea8cdd6de3f553/ichnaea/log.py#L275-L309

That's the three arguments possibility for check. So `self.metric_type + ".result"` is the name, `"fallback_allowed:false"` is the match, and the value is the list of tags. The code then checks to see if the value is a list and if so, that becomes the tags and the value becomes `None`.

The match gets ignored because it's not an int. Seems like this is a bug. There are a couple of other instances of this--but not many in the grand scheme of things.

Anyhow, `.check()` checks to see if the listed tags are a subset of the tags of the metrics message. In this case, it's a subset, so `.check()` works fine.

However, Markus checks to see if the expected list of tags is exactly the same as the actual list of tags and it's not because `"fallback_allowed:false"` isn't in the expected list. So I added that and now the test passes.

Why write all this out? Because it's not straight-forward and it was curious and while I think I got it right, maybe I didn't.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -522,38 +523,44 @@ def _query(self, **kwargs):
...
         return Query(**kwargs)
 
-    def test_get_blue(self, cache, stats):
+    def test_get_blue(self, cache, metricsmock):
         blues = BlueShardFactory.build_batch(2)
         query = self._query(blue=self.blue_model_query(blues))
         assert cache.get(query) is None
-        stats.check(
-            counter=[
-                ("locate.fallback.cache", 1, 1, [self.fallback_tag, "status:miss"])
```
willkg: This has four values and they get unpacked as `(name, match, value, tags)` where "match" is the number of records that match those criteria. I'm not sure it matters here whether there's one record or more than one, so I didn't add that assertion.

Now I feel bad about that decision. I'll re-add it.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -762,12 +797,25 @@ def test_success(self, geoip_db, http_session, session, source, stats):
...
+        assert metricsmock.has_record(
+            "incr",
+            "locate.fallback.cache",
+            value=1,
+            tags=[self.fallback_tag, "status:miss"],
+        )
+        assert metricsmock.has_record(
+            "timing", "locate.fallback.lookup", tags=[self.fallback_tag]
+        )
+        assert metricsmock.has_record(
```
willkg: I think this one is wrong. I don't think this should have gotten added.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -778,44 +826,35 @@ def test_cache_empty_result(self, geoip_db, http_session, session, source, stats
...
-                            self.fallback_tag,
-                            "status:%s" % self.fallback_not_found_status,
-                        ],
-                    ),
-                ]
+            assert metricsmock.has_record(
+                "incr",
+                "locate.fallback.cache",
+                value=1,
+                tags=[self.fallback_tag, "status:hit"],
```
willkg: This is missing a check.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -1145,34 +1168,58 @@ def test_set_cache_redis_failure(
...
 
             assert mock_request.call_count == 1
-            stats.check(
-                counter=[
-                    ("locate.fallback.cache", [self.fallback_tag, "status:miss"]),
-                    ("locate.fallback.lookup", [self.fallback_tag, "status:200"]),
-                ],
-                timer=[("locate.fallback.lookup", [self.fallback_tag])],
-            )
+            assert metricsmock.get_records() == [
```
willkg: I'll switch this to individual checks since it'll be less fragile.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -1145,34 +1168,58 @@ def test_set_cache_redis_failure(
...
+                    ["fallback_name:fall", "status:miss"],
+                ),
+                ("timing", "locate.fallback.lookup", mock.ANY, ["fallback_name:fall"]),
+                (
+                    "incr",
+                    "locate.fallback.lookup",
+                    1,
+                    ["fallback_name:fall", "status:200"],
+                ),
+                (
```
willkg: This last check shouldn't be here.

ichnaea/api/locate/tests/test_fallback.py
```diff
@@ -1181,15 +1228,28 @@ def test_cache_single_cell(self, geoip_db, http_session, session, source, stats)
...
 
             assert mock_request.call_count == 1
-            stats.check(
-                counter=[
-                    ("locate.fallback.cache", [self.fallback_tag, "status:hit"]),
-                    ("locate.fallback.lookup", [self.fallback_tag, "status:200"]),
-                ],
-                timer=[("locate.fallback.lookup", [self.fallback_tag])],
-            )
+            assert metricsmock.get_records() == [
```
willkg: I'll switch this to individual checks, too.

ichnaea/api/locate/tests/test_fallback.py
willkg: My bad--this is a bug in the tests. The statsd state isn't getting cleared, so the second metric check is actually from the first set. When I add the `metricsmock.clear_records()`, then the second metric isn't there anymore, so we shouldn't be checking for it.

ichnaea/api/locate/tests/test_query.py
```diff
@@ -352,63 +343,58 @@ def _make_query(
...
-            counter=[
-                (
-                    "locate.query",
-                    ["key:test", "region:GB", "blue:one", "cell:one", "wifi:one"],
-                )
-            ],
-        )
+            geoip_db, blue=blues, cell=cells, wifi=wifis, ip=self.london_ip
+        )
+        assert metricsmock.get_records() == [
```
willkg: Since the previous check makes sure there's only one of these `incr` records, I converted it to this `get_records()` assertion which both asserts that there's only one record and it looks like this.

ichnaea/api/submit/tests/base.py
```diff
@@ -100,51 +100,68 @@ def test_error_redis_failure(self, app, raven, stats):
...
         assert redis.keys("apiuser:*") == []
 
-    def test_log_stats(self, app, redis, stats):
+    def test_log_stats(self, app, redis, metricsmock):
         cell, query = self._one_cell_query()
         self._post(app, [query], api_key="test")
-        stats.check(
-            counter=[
-                ("data.batch.upload", 1),
-                ("data.batch.upload", ["key:test"]),
```
willkg: These two check the same record.

ichnaea/conftest.py
```diff
@@ -303,38 +304,47 @@ def redis(redis_client):
...
+                'incr',
+                stat='some.stat',
+                value=1
+            )
+
+    If you ever need to clear the records in the middle of a test, do::
+
+        metricsmock.clear_records()
+
+    See for more: https://markus.readthedocs.io/en/latest/testing.html
```
willkg: This is awkward phrasing.

ichnaea/data/tests/test_export.py
```diff
@@ -265,12 +265,32 @@ def test_upload(self, celery, session, stats):
         gotten = [report["position"]["accuracy"] for report in send_reports]
         assert set(expect) == set(gotten)
 
-        stats.check(
-            counter=[
-                ("data.export.batch", 4, 1, ["key:backup"]),
```
willkg: The 4 here means that it's checking for 4 separate records like this. I converted this to a `len` of a `filter_records()` call.

ichnaea/data/tests/test_export.py
```diff
@@ -452,31 +472,32 @@ def test_stats(self, celery, session, stats):
...
-        for name, total in (("cell", 12), ("wifi", 24)):
-            insert_msgs = [
-                msg
-                for msg in stats.msgs
-                if (msg.startswith("data.observation.insert") and "type:" + name in msg)
-            ]
-            assert (
-                sum([int(msg.split(":")[1].split("|")[0]) for msg in insert_msgs])
-                == total
-            )
```
willkg: The way I rewrote this covers both sets of assertions. It asserts the batch and upload messages are covered and by which keys. It also asserts the total number of inserted items by type.

ichnaea/data/tests/test_export.py
```diff
@@ -558,16 +579,18 @@ def test_cell_duplicated(self, celery, session):
...
-                ("data.observation.drop", 1, 1, ["type:cell", "key:test"]),
-            ]
+        assert metricsmock.has_record(
+            "incr", "data.report.upload", value=1, tags=["key:test"]
+        )
+        assert metricsmock.has_record(
+            "incr", "data.report.drop", value=1, tags=["key:test"]
+        )
+        assert metricsmock.has_record(
+            "incr", "data.observation.drop", value=1, tags=["type:cell", "key:test"]
```
willkg: These don't assert that there's only one of each of these records. I'll fix that.

ichnaea/data/tests/test_station.py
```diff
@@ -101,29 +103,32 @@ def mock_area(self, updated_areas, key, num=num, session2=session2):
...
-            self.check_statcounter(redis, StatKey.cell, 1)
-            self.check_statcounter(redis, StatKey.unique_cell, 1)
-            stats.check(
-                counter=[("data.observation.insert", 1, ["type:cell"])],
-                timer=[("task", 1, ["task:data.update_cell"])],
-            )
-        finally:
-            CellUpdater._retry_wait = orig_wait
-            session.execute(text("drop table %s;" % cell.__tablename__))
+        with metricsmock as mm:
```
willkg: Oops--this is leftover from my first iteration. This shouldn't be used in a context manager anymore.

ichnaea/data/tests/test_station.py
```diff
@@ -101,29 +103,32 @@ def mock_area(self, updated_areas, key, num=num, session2=session2):
...
-            shard = CellShard.shard_model(obs.cellid)
-            cells = session.query(shard).all()
-            assert len(cells) == 1
-            assert cells[0].samples == 1
-
-            self.check_statcounter(redis, StatKey.cell, 1)
-            self.check_statcounter(redis, StatKey.unique_cell, 1)
-            stats.check(
-                counter=[("data.observation.insert", 1, ["type:cell"])],
-                timer=[("task", 1, ["task:data.update_cell"])],
```
willkg: These assert there's only one of each of these messages. I need to re-implement that.

ichnaea/data/tests/test_station.py
```diff
@@ -240,45 +245,47 @@ def test_blocklist_skip(self, celery, redis, session):
...
-            assert station.samples == 5
-            assert station.source == source
-            assert round(station.weight, 2) == 5.0
-            self.check_blocked(station, None)
-            self.check_dates(station, self.today, self.today, self.today)
-
-        stats.check(
-            counter=[
-                ("data.observation.insert", 2, [self.type_tag]),
-                ("data.station.new", 2, [self.type_tag]),
```
willkg: These assert that there are two of each of these messages. I need to reimplement that.

ichnaea/log.py
```diff
@@ -59,6 +62,15 @@ def configure_logging():
         "root": {"handlers": handlers, "level": "WARNING"},
     }
 
+    if local_dev_env:
+        # In the local dev environment, add markus as a logger so that we can
+        # see metrics in the logs
+        logging_config["loggers"]["markus"] = {
+            "propagate": False,
+            "handlers": handlers,
+            "level": level,
```
willkg: Markus doesn't log anything unless the logging backend is configured, so I think I'll tweak this so it's not optional.

willkg:
^^^ I reviewed the whole thing and fixed a bunch of issues. It's good to review now. Sorry it's so long.

ichnaea/api/locate/tests/base.py
jwhitlock: I think this was a mistake as well. It was added in 2015 in https://github.com/mozilla/ichnaea/commit/add1bacceff77b0cd4706590a42bea6fb5dddfef, and there are many similar tests that got updated to add ``fallback_allowed:false``. Most correctly add to the tag list, but there are a sequence of them in this file where it got added in the wrong place.  It is a shame that there were no assertions for a missing tag in the list, or that ``match`` was a non integer. Glad to see this fixed!

ichnaea/api/locate/fallback.py
```diff
@@ -597,11 +596,12 @@ def _make_external_call(self, query, fallback_schema):
...
             fallback_tag = "fallback_name:%s" % (query.api_key.fallback_name or "none")
 
-            with self._stat_timed("lookup", tags=[fallback_tag]):
+            with METRICS.timer("locate.fallback.lookup", tags=[fallback_tag]):
                 response = outbound_call(query, outbound)
 
-            self._stat_count(
-                "lookup", tags=[fallback_tag, "status:" + str(response.status_code)]
+            METRICS.incr(
+                "locate.fallback.lookup",
```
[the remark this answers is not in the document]
willkg: Oh, the key shouldn't be the same for incr and timing. I didn't catch that.

As an aside, this is something I've been trying to figure out how to do with Markus: how can you register all the metrics keys in use such that you can:

1. make sure you're not using the same key for different metrics types (incr, gauge, etc)
2. auto-document all the keys in use for external documentation
3. reduce likelihood of mistyped keys

It's tricky.

ichnaea/log.py
```diff
@@ -91,21 +94,29 @@ def configure_raven(transport=None, _client=None):
...
-    client = klass(host=statsd_host, port=statsd_port, namespace=namespace, use_ms=True)
-    return client
+def configure_stats():
+    """Configure Markus for metrics."""
+    local_dev_env = settings("local_dev_env")
+    if local_dev_env:
+        markus.configure(backends=[{"class": "markus.backends.logging.LoggingMetrics"}])
+        return
+
+    if settings("statsd_port"):
```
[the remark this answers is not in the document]
willkg: Oops--I'll fix this.

ichnaea/api/locate/tests/base.py
willkg: Oh, that's brilliant! I'll look at what other mock assertions make sense to have. It's very much like assertions on function calls.

willkg:
That fixes the `statsd_port` and `locate.fallback.lookup` issues. I think I'm going to push the markus API changes to a future PR since we need to do some development in markus for that.

willkg:
Thank you!

PR #948, opened 2019-10-09
Add field definitions for /v1/submit/ API docs

willkg:
When I removed the deprecated `/v1/search/` API endpoint code and documentation, I didn't realize the documentation for the `/v1/submit/` API endpoint referred to it for the field definitions.

This adds the field definitions to the `/v1/submit/` API endpoint.

### DIFF
--- a/docs/api/submit.rst
+++ b/docs/api/submit.rst
@@ -73,9 +73,6 @@ Here is an example position report:
 Field Definition
 ================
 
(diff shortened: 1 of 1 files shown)


willkg:
I copied the old docs into the place that refers to them in the new docs. Since this is a deprecated endpoint, I'm going to self-r+ this PR as "better". If we want to hone it further, we can do that in future PRs.

PR #949, opened 2019-10-11
Update docs regarding change in locate.fallback.lookup* metrics

willkg:
This fixes the docs after changes made in #944.


willkg:
self-r+

Issue #921, opened 2019-09-23
test_locate_v1.test_blue_seen fails occasionally


jwhitlock:
The test ``test_blue_seen`` in ``ichnaea/api/locate/tests/test_locate_v1.py`` fails occasionally, such as a recent failure:

```
___________________________ TestView.test_blue_seen ____________________________
Traceback (most recent call last):
  File "/app/ichnaea/api/locate/tests/test_locate_v1.py", line 176, in test_blue_seen
    self.check_queue(data_queues, 0)
  File "/app/ichnaea/api/locate/tests/base.py", line 210, in check_queue
    assert data_queues["update_incoming"].size() == num
AssertionError
```

As [willkg says](https://github.com/mozilla/ichnaea/pull/920#issuecomment-534136454), it's weird that the assertion error shows like that and doesn't show the pytest fancy assertion stuff.

Here's the code:

https://github.com/mozilla/ichnaea/blob/7ee9de14a2b44456adf71b3d3113f5cbdf851bdf/ichnaea/api/locate/tests/test_locate_v1.py#L168-L176

The "v1" API is the current locate API, "v0" is the deprecated API. ``BlueShardFactory`` has some randomness for the MAC address, but the lat and long are constant:

https://github.com/mozilla/ichnaea/blob/5b020c4bd41246f3cefb98949a9555e598fe7e45/ichnaea/tests/factories.py#L143-L157

The frequency of failures may give some information, as well as if it is always this test that fails.

jwhitlock:
It failed today on [build 1703](https://circleci.com/gh/mozilla/ichnaea/1703), for the PR. It failed 4 days ago in [build 1689](https://circleci.com/gh/mozilla/ichnaea/1689), merging a PR to master.  That's 14 builds between failures, so if it was pure randomness, I'd expect to have seen more failures in the previous month. This makes it more likely to be a recent requirements update 😓

jwhitlock:
Recent failures:

* Wed Oct 2 9:13 AM: [Build 1756](https://circleci.com/gh/mozilla/ichnaea/1756)
* Wed Oct 9 8:22 AM: [Build 1783](https://circleci.com/gh/mozilla/ichnaea/1783)

Weekly is still super annoying. I'm starting with investigating why the pytest enchanced assert isn't being used.

jwhitlock:
With PR #950 merged, we should get pytest enhanced ``assert`` the next this fails.

jwhitlock:
[...]

PR #950, opened 2019-10-14
Investigate occasional test failure in the locate API

jwhitlock:
Issue #921 tracks a test that occasionally fails (1-2 times a week) with no relevant code changes. This PR makes two changes:

* Use ``pytest`` rich ``assert`` in ``ichnaea/api/locate/tests/base.py`` and ``ichnaea/api/submit/tests/base.py``. This will tell us how many items have been added to the ``update_incoming``, which is expected to be empty. The shared testing code in these files has been skipped because they are named ``base.py`` rather than ``test_base.py`` or ``base_test.py``. This is a smaller change than renaming the files, and avoids confusion with ``ichnaea/models/tests/test_base.py``.
* Confirm that the data queue is empty at the start of the test, so that we know whether there is a test setup issue or the API view is queuing the report for further processing.

### DIFF
--- a/ichnaea/api/submit/tests/__init__.py
+++ b/ichnaea/api/submit/tests/__init__.py
@@ -1,4 +1,4 @@
 import pytest
 
 # Use pytest rich asserts in a file that doesn't match test_* pattern
-pytest.register_assert_rewrite("ichnaea.api.locate.tests.base")
+pytest.register_assert_rewrite("ichnaea.api.submit.tests.base")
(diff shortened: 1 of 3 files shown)


ichnaea/api/submit/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+import pytest
+
+# Use pytest rich asserts in a file that doesn't match test_* pattern
+pytest.register_assert_rewrite("ichnaea.api.locate.tests.base")
```
willkg: I think this one needs to be `ichnaea.api.submit.tests.base`.

jwhitlock: 🤦‍♂ yep copy and paste gets me again

PR #951, opened 2019-10-14
Re-add monitor queue size task

### LINKED ISSUE #947: monitor data queue size
There are a series of celery tasks that are tied to data queues. If the tasks aren't processing items in the data queues fast enough, the data queues can back up infinitely until Redis evicts them to reclaim memory.

We need monitoring on the data queue sizes so we can adjust Celery processing.

willkg:
This re-adds the monitor queue size task. As MLS handles larger loads, the data queues can back up and if we're not able to see them, we can't do anything about it.

Fixes #947.

### DIFF
--- a/ichnaea/data/monitor.py
+++ b/ichnaea/data/monitor.py
@@ -60,3 +60,22 @@ class ApiUsers(object):
                 value=value,
                 tags=["key:%s" % api_name, "interval:%s" % interval],
             )
+
+
+class QueueSize(object):
+    """Generate gauge metrics for all queue sizes.
+
+    This covers the export queues, the celery task queues, and the
+    data queues.
+
+    """
+
+    def __init__(self, task):
+        self.task = task
+
+    def __call__(self):
+        keys = self.task.redis_client.scan_iter(match="export_queue_*", count=100)
+        export_queues = set([key.decode("utf-8") for key in keys])
+        for name in export_queues | self.task.app.all_queues:
+            value = self.task.redis_client.llen(name)
+            METRICS.gauge("queue", value, tags=["queue:" + name])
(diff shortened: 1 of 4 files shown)


ichnaea/data/monitor.py
```diff
@@ -60,3 +60,22 @@ def __call__(self):
...
+
+    def __init__(self, task):
+        self.task = task
+
+    def __call__(self):
+        keys = self.task.redis_client.scan_iter(match="export_queue_*", count=100)
+        export_queues = set([key.decode("utf-8") for key in keys])
+        for name in export_queues | self.task.app.all_queues:
+            value = self.task.redis_client.llen(name)
+            METRICS.gauge("queue", value, tags=["queue:" + name])
```
willkg: This is pretty much the old monitor code--I just re-added it.

ichnaea/taskapp/config.py
```diff
@@ -102,11 +102,10 @@ def init_worker(
...
     celery_app.geoip_db = configure_geoip(raven_client=raven_client, _client=_geoip_db)
 
-    # configure data queues
-    celery_app.all_queues = all_queues = set([q.name for q in TASK_QUEUES])
-
+    # configure data queues and build set of all queues
+    all_queues = set([q.name for q in TASK_QUEUES])
     celery_app.data_queues = data_queues = configure_data(redis_client)
-    all_queues = all_queues.union(
+    celery_app.all_queues = all_queues.union(
```
willkg: This is the money fix. Previously, `all_queues` would end up with a set of all the queues: task and data and export. However, since that wasn't persisted anywhere, it'd get lost. `celery.all_queues` was only the task queues.

If we do this same fix in the 1.5 branch, we'll be able to see metrics on the data queues. I'll do that in a followup PR.

willkg:
Thank you!

I'll do a PR for the 1.5 branch now.

PR #952, opened 2019-10-14
Fix monitoring data queue sizes

### LINKED ISSUE #947: monitor data queue size

willkg:
This fixes the `celery_app.all_queues` set so it includes the data queues. That way data queue sizes are sent as gauge metrics.

Fixes #947

### DIFF
--- a/ichnaea/async/config.py
+++ b/ichnaea/async/config.py
@@ -151,10 +151,9 @@ def init_worker(celery_app,
         raven_client=raven_client, _client=_geoip_db)
 
     # configure data queues
-    celery_app.all_queues = all_queues = set([q.name for q in CELERY_QUEUES])
-
+    all_queues = set([q.name for q in CELERY_QUEUES])
     celery_app.data_queues = data_queues = configure_data(redis_client)
-    all_queues = all_queues.union(
+    celery_app.all_queues = all_queues.union(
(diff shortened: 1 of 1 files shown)


ichnaea/async/config.py
```diff
@@ -151,10 +151,9 @@ def init_worker(celery_app,
         raven_client=raven_client, _client=_geoip_db)
 
     # configure data queues
-    celery_app.all_queues = all_queues = set([q.name for q in CELERY_QUEUES])
-
+    all_queues = set([q.name for q in CELERY_QUEUES])
     celery_app.data_queues = data_queues = configure_data(redis_client)
-    all_queues = all_queues.union(
+    celery_app.all_queues = all_queues.union(
```
willkg: The monitor queues task looks at `celery_app.all_queues`. This fixes that set so it's the set of all the task queues and the data queues. Previously, `all_queues` had the task and data queues, but `celery_app.all_queues` was just the task queues.

willkg:
Tomorrow, I'll make sure this gets to stage and prod and test it out.

PR #954, opened 2019-10-15
Test, fix cell area region algorithm when cell regions are None

jwhitlock:
When importing the full cell export, a ``TypeError`` is raised when creating a cell area for some cell groups, due to changes in Python 3.x around comparing to ``None``. This PR avoids the ``TypeError`` and makes the code work like it did in Python 2.6.

In Python 2.6, you can sort ``None`` versus strings, and ``None`` is always less than a string:

```python
>>> sorted(['c', 'a', 'b', '', None])
[None, '', 'a', 'b', 'c']
```

In Python 3.7, it is a ``TypeError`` to compare ``None`` and other values:
```python
>>> sorted(['c', 'a', 'b', '', None])
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'NoneType' and 'str'
```

This can happen when the ``CellAreaUpdater`` is trying to determine the region for a group of cells with mixed regions, which may include ``None`` in cases where the geocoder can't determine the region. This happens in the dev environment for certain islands in Europe, such as [Sardinia](https://en.m.wikipedia.org/wiki/Sardinia) (a region of Italy) or [Bornholm](https://en.m.wikipedia.org/wiki/Bornholm) (a region of Denmark). It is possible the licensed MaxMindDB is more detailed, but this would still happen in the case of cell towers erroneously located in the ocean. 

This PR adds a test for all the cell regions being ``None`` (the cell area region is also ``None``), and if there is a tie between ``None`` and another region. This second test fails, raising a ``TypeError``, until the code changes that set the cell area region to ``None``, in the same way as Python 2.x sorting would work.

### DIFF
--- a/ichnaea/data/area.py
+++ b/ichnaea/data/area.py
@@ -44,7 +44,9 @@ class CellAreaUpdater(object):
                 grouped_regions[reg] += 1
             max_count = max(grouped_regions.values())
             max_regions = sorted(
-                [k for k, v in grouped_regions.items() if v == max_count]
+                [k for k, v in grouped_regions.items() if v == max_count],
+                # Emulate Python 2.6, where None is sortable and first
+                key=lambda item: item or "",
             )
             # If we get a tie here, randomly choose the first.
             region = max_regions[0]
--- a/ichnaea/data/tests/test_area.py
+++ b/ichnaea/data/tests/test_area.py
@@ -1,5 +1,7 @@
 from datetime import timedelta
 
+import pytest
+
 from ichnaea.data.tasks import update_cellarea
 from ichnaea.models import area_id, encode_cellarea, CellArea, Radio
 from ichnaea.tests.factories import CellAreaFactory, CellShardFactory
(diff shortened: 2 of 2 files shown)


ichnaea/data/area.py
```diff
@@ -43,11 +43,13 @@ def region(self, ctr_lat, ctr_lon, mcc, cells):
...
-                [k for k, v in grouped_regions.items() if v == max_count]
-            )
+            max_regions = [k for k, v in grouped_regions.items() if v == max_count]
             # If we get a tie here, randomly choose the first.
-            region = max_regions[0]
+            if None in max_regions:
+                # Emulate Python 2.6, where None is sorted first
+                region = None
+            else:
+                region = sorted(max_regions)[0]
```
willkg: Why not use a key? Something like this:

```
region = sorted(max_regions, key=lambda item: "" if region is None else region)
```

willkg: Or more korrect:

```
region = sorted(max_regions, key=lambda item: item or "")
```

jwhitlock: I feel weird using ``key`` to work around this change, but it does make it a one-line change, and could be use other places in the code where ``sorted``/``sort`` breaks unexpectedly.

jwhitlock:
Force-push to use ``key=lambda item: item or ""``

PR #959, opened 2019-10-16
Add documentation for SQLALCHEMY_URL

### LINKED ISSUE #904: nix ddl uri

willkg:
Fixes #904

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -20,6 +20,14 @@ Environment variables
    :case: upper
 
 
+Alembic requires an additional item in the environment:
+
+.. code-block:: bash
+
+   # URI for user with ddl access
+   SQLALCHEMY_URL=mysql+pymysql://USER:PASSWORD@HOST:PORT/DBNAME
+
+
 The webapp uses gunicorn which also has configuration.
 
 .. literalinclude:: ../docker/run_web.sh


docs/configure.rst
```diff
@@ -20,6 +20,14 @@ Environment variables
    :case: upper
 
 
+Alembic requires an additional item in the environment:
+
+.. code-block:: bash
+
+   # URI for user with ddl access
+   SQLALCHEMY_URL=mysql+pymysql://USER:PASSWORD@HOST:PORT/DBNAME
```
willkg: I verified the variable is `SQLALCHEMY_URL` as opposed to `...URI`.

I also verified the structure is correct.

Having said that, the user doesn't need ddl access, so I'll fix the comment.

willkg:
`DB_READONLY_URI` and `DB_READWRITE_URI` are for configuring sqlalchemy in the webapp and celery. Specifically, the webapp always uses the readonly one and celery uses the readwrite one.

Alembic has a `sqlalchemy_url` configuration value. The `SQLALCHEMY_URL` environment variable is explicitly for setting that.

There's no unification required here unless we plan to ditch the mysql replica and run everything off of one mysql server using a user account that has complete access to the data and making table changes. Pretty sure we don't want to do that at this time.

willkg:
Thank you for reviewing! Sorry I missed this when I did the initial `SQLALCHEMY_URL` work.

Issue #960, opened 2019-10-17
Split metrics being used as counters and timers


jwhitlock:
There are some metrics that are being used as both counters and timers. This is confusing, and possibly invalid.

``locate.fallback.lookup`` was split into ``locate.fallback.lookup_timing`` and ``locate.fallback.lookup_count`` in PR #944. 

``data.export.upload`` is also used as both a timer and a counter.

I'll add more as I find them.

jwhitlock:
The ``request`` metric is also used as a timer and counter. 

I've gone through the current metrics code and haven't found any more dual-use metrics.

hannosch:
The metrics code started out with graphite as a backend, where the metric type was an enforced top level namespace. So repeating the type inside the metric name was redundant, see for example https://github.com/lukevenediger/statsd.net/blob/master/statsd.net/Documentation/guidance/metric-anti-patterns.md#anti-pattern-redundant-endings-such-as-timing-count-or-total

IIRC datadog instead added the type as an automatic postfix, so you had things like ``metric.name.count``, ``metric.name.avg``, ``metric.name.min`` etc. where the min/avg/max postfix represented the timer. Repeating the metric type would have resulted in silly looking ``metric.name_count.count``.

The statsd protocol itself doesn’t mention this topic at all.

jwhitlock:
Thanks @hannosch, that helps.

We're using InfluxDB to store metrics, and Grafana to explore and display them. I looked at ``locate.fallback.lookup`` from a few months ago, when we were using fallback lookups. The only datapoints I was able to see had the tag ``metric_type=counter``, and there was no data with the tag ``metric_type=timing``, despite [the code](https://github.com/mozilla/ichnaea/blob/446999ea1ce10c8a4627ffed733cd25c9dc12c75/ichnaea/api/locate/fallback.py#L360-L370) recording both a counter and a timer under that name. I was able to see ``metric_type=timing`` data in ``datamaps``, which only is being sent timer data. It appears to store it as milliseconds, as an integer.

I can't determine from the InfluxDB docs why there aren't records for both the timer and counter for ``locate.fallback.lookup`` or the other dual-use metrics. This could be a bug in InfluxDB, or the metrics client in our code, or even a limitation of Grafana. In any case, it would be easy to avoid by sending only one metric type per name.

I see two ways to proceed. One way is to add `.timing` to timer metric names:

* ``locate.fallback.lookup_count`` back to ``locate.fallback.lookup``
[...]

PR #961, opened 2019-10-17
Add table of contents to metrics doc

jwhitlock:
When exploring metrics in Grafana, it is helpful to go from an alphabetical list of metric names to the detailed documentation for that metric.

I was expecting to find that the document was out of date, but it is actually a good reflection of the metrics code. There are no documented metrics that aren't omitted, and no emitted metrics that aren't documented. I left the note ``2019-09-24: This needs to be updated``, but I believe the docs are accurate, and I'd support removing it.

I'd want to rewrite this documentation. I don't find the purpose-based organization to be very helpful, and I'd like to get away from the notation ``key#tag1=value1,tag2=value2``. I'm holding back on a rewrite because 1) we should focus on the new infrastructure work, and 2) @willkg has been thinking of ways to define the tags and documentation in the code and auto-generate docs.

I did discover some dual-use metrics, which I think should be split into counters and timers. They show up twice in the table, and issue #960 can be used to discuss what we should do to split them up.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -6,17 +6,55 @@ Metrics
 
 .. Note:: 2019-09-24: This needs to be updated.
 
-As discussed in :ref:`the deployment document <deploy>`, Ichnaea emits
-metrics through the StatsD client library with the intent of
-aggregating and viewing them on a compatible dashboard.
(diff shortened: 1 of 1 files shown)


jwhitlock:
Thanks!

PR #962, opened 2019-10-21
Split counter and timer metrics

jwhitlock:
This PR creates new metrics for timing values, so that we are not using the same metric name for a counter and a timer:

* ``data.export.upload`` → ``data.export.upload.timing``
* ``request`` → ``request.timing``

The counters continue sending to the original names, ``data.export.upload`` and ``request``.

This is the fix to #960 that requires the least code changes. For consistency, I also changed two metrics to use the same naming convention:

* ``locate.fallback.lookup_timing`` → ``locate.fallback.lookup.timing``
* ``locate.fallback.lookup_count`` → ``locate.fallback.lookup``

This is not the only possible naming scheme, and I'm open to suggestions for other ones.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -25,7 +25,7 @@ Metric Name                      Type    Tags
 `data.batch.upload`_             counter key
 `data.export.batch`_             counter key
 `data.export.upload`_            counter key, status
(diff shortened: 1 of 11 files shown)


jwhitlock:
Thanks!

Issue #967, opened 2019-10-21
improve debugging startup issues in webapp


willkg:
If there's a configuration issue, then the webapp will mysteriously refuse to start, but there's no logs or errors or any other indication of what's going on.

We added a `LOCAL_DEV_ENV` mode for running the webapp and using this can sort of help.

Maybe it'd be better to have the `docker/run_web.sh` script check the webapp before trying to run it as a webapp?

willkg:
I think the changes I did in PR #968 cover this pretty well. Anything more seems like diminishing returns. Going to close this out.

PR #968, opened 2019-10-21
Improve webapp startup debugging

willkg:
This improves the ability to debug the webapp startup in server
environments where startup issues are masked by gunicorn running the
webapp in a separate worker process. Now when the webapp container is
started, it runs a "check" to see if things can start up, and then
proceeds to run the webapp in gunicorn.

This helps fix issue #967.

### DIFF
--- a/ichnaea/webapp/app.py
+++ b/ichnaea/webapp/app.py
@@ -1,8 +1,15 @@
 """
 Holds global web application state and the WSGI handler.
+
+You can run this script for a one-process webapp.
+
+Further, you can pass in ``--check`` which will create the app and then exit
+making it easier to suss out startup and configuration issues.
+
 """
 
 import logging
+import sys
 
 from waitress import serve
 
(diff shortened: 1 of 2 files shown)


ichnaea/webapp/app.py
```diff
@@ -66,4 +73,9 @@ def log_response(status, headers, exc_info=None):
 
 
 if __name__ == "__main__":
-    serve(log_access_factory(main(ping_connections=True)), host="0.0.0.0", port=8000)
+    if "--check" in sys.argv:
+        main(ping_connections=False)
```
willkg: This is a pretty benign mode, however it makes sure that the webapp loads and it'll suss out most configuration issues. I say "most" because configuration throws errors "on demand" and this app doesn't log configuration, so there are a couple of configuration variables that throw an error during the http request/response time. However, that will send an error to sentry and it's not at startup, so that's "ok".

jwhitlock:
Can you recall one of the errors that was hard to debug before?

jwhitlock:
I figured out my problem by reading issue #967. Adding ``LOCAL_DEV_ENV=False`` to ``my.env`` caused ``gunicorn`` to fail silently, while running ``--check`` continues to fail loudly 👍

willkg:
One of the problems was when setting `REDIS_URI` to a host with no `redis://` at the beginning.

I just pushed a fix so this works from the shell now.

After the fix, if you do:

```
$ make shell
app@...:/app$ REDIS_URI=example.com /app/docker/run_web.sh
```

it'll throw an obvious error now. Before, it'd fail in gunicorn, but it was unclear why.

It's a lousy error for this specific situation. I've been mulling over what to do about that. That's something we can fix in a future PR.

willkg:
Thank you!

Issue #963, opened 2019-10-21
verify sentry is working


willkg:
This covers verifying that sentry is working in the new infrastructure.

1. is the sentry configuration correct and pointing to a valid sentry project that we can send events to?
2. can the admin node send events to sentry?
3. can the webapp send events to sentry?
4. can the celery workers send events to sentry?

willkg:
Also, we should make sure that sentry is getting the git sha of Ichnaea in the tags.

willkg:
Once that gets deployed, we can run the `clitest` and `celerytest` and verify sentry is working for the admin node and celery workers. That covers questions 1, 2, and 3.

@jwhitlock and I think we're going to punt on figuring out the webapp (question 4). There are ways to do it, but they're all gross.

willkg:
We pushed out master tip to stage.

I ran the `clitest` and it did fine so configuration for the sentry project is correct:

https://sentry.prod.mozaws.net/operations/location-stage/issues/6556132/?query=is:unresolved

Further, that has the right release value now.

I ran the celery test which creates a celery task and queues it up. When the task is worked on, it sends a message to celery. Nothing made it to celery. The logs have this:

```
msg: "Received unregistered task of type 'ichnaea.data.tasks.sentry_test'.
The message has been ignored and discarded.

Did you remember to import the module containing this task?
Or maybe you're using relative imports?

Please see
http://docs.celeryq.org/en/latest/internals/protocol.html
for more information.

The full contents of the message body was:
b'[[], {"msg": "Testing Sentry configuration via celery (2019-10-28 16:45:16.865092)"}, {"callbacks": null, "errbacks": null, "chain": null, "chord": null}]' (154b)
"    
   traceback: "Uncaught exception:
  File "/usr/local/lib/python3.7/site-packages/celery/worker/consumer/consumer.py", line 559, in on_task_received
    strategy = strategies[type_]
<class 'KeyError'>
KeyError('ichnaea.data.tasks.sentry_test')
```

Looking into this today.

willkg:
Here's another test:

* cli: https://sentry.prod.mozaws.net/operations/location-stage/issues/6601457/
* celery: https://sentry.prod.mozaws.net/operations/location-stage/issues/6601458/

We can't explicitly test the webapp. We'll have to wait until it throws an error.

willkg:
We verified that sentry is working for cli, celery, and webapp in both stage and prod. We're good here. Marking as FIXED.

PR #969, opened 2019-10-22
Fix sentry reporting to use git sha for release

willkg:
If there's no tag, use the git commit as the release in sentry
reporting. This makes it easier to figure out bugs when we're
auto-deploying to stage when things land in the master branch.

Part of issue #963.

### DIFF
--- a/ichnaea/log.py
+++ b/ichnaea/log.py
@@ -89,7 +89,8 @@ def configure_raven(transport=None, _client=None):
 
     dsn = settings("sentry_dsn")
     klass = DebugRavenClient if not dsn else RavenClient
-    release = version_info()["tag"]
+    info = version_info()
+    release = info.get("version") or info.get("commit") or "unknown"
     client = klass(dsn=dsn, transport=transport, release=release)
     return client
 
--- a/version.json
+++ b/version.json
@@ -1,7 +1,6 @@
 {
-    "build": null,
-    "commit": "HEAD",
     "source": "https://github.com/mozilla/ichnaea",
-    "tag": "master",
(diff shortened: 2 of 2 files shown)


ichnaea/log.py
```diff
@@ -89,7 +89,8 @@ def configure_raven(transport=None, _client=None):
 
     dsn = settings("sentry_dsn")
     klass = DebugRavenClient if not dsn else RavenClient
-    release = version_info()["tag"]
+    info = version_info()
+    release = info.get("tag") or info.get("commit") or "unknown"
```
willkg: Oh, whoops... this isn't right. This works with the `version.json` that's in the repo, but that doesn't match the one that's built by `.circleci/config.yml`.

Mmm... I'll fix all of them.

willkg:
^^^ That fixes `version.json` to match what we're creating in `.circleci/config.yml` and the logic in the sentry code to set the `release` thing.

version.json
```diff
@@ -1,7 +1,6 @@
 {
-    "build": null,
-    "commit": "HEAD",
     "source": "https://github.com/mozilla/ichnaea",
-    "tag": "master",
-    "version": "dev"
+    "version": "unknown",
+    "commit": "HEAD",
+    "build": "unknown"
```
jwhitlock: What is this new key order?

willkg: It matches what's in the spec. Figured it's easier to eyeball:

https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md

willkg:
Thank you!

PR #973, opened 2019-10-24
Add sentry_test.py script

willkg:
This adds a sentry_test.py script that can run two different tests:

1. `clitest` which verifies the Sentry configuration and whether the
   admin node can send errors to Sentry
2. `celerytest` which queues a Celery task to send a message which
   tests whether Celery tasks can send errors to Sentry

Issue #963


willkg:
We can test both of these with `DebugRavenClient` in the local dev environment.

1. run `make shell` in one terminal and `make runcelery` in another terminal
2. run `python ichnaea/scripts/sentry_test.py clitest` in the shell
   It spits out some messages, but doesn't do much because `DebugRavenClient` doesn't print anything when it's sending stuff.
3. run `python ichnaea/scripts/sentry_test.py celerytest` in the shell
   You can see the task get queued and execute in Celery.

I tested the `clitest` subcommand on the admin node in stage. It printed this to the console:

```
app@xxx:/app$ python ichnaea/scripts/sentry_test.py clitest
Using message: Testing Sentry configuration via cli (2019-10-24 16:01:05.196788)
Building Raven client...
Sending message...
app@xxx:/app$ 
```

That resulted in this Sentry item:

https://sentry.prod.mozaws.net/operations/location-stage/issues/6529328/?query=is:unresolved

I like how there's a timestamp printed to the console and the Sentry issue so you can correlate the two even if you've done a bunch of testing. The celery test is similar.

I can't test `celerytest` on the admin node until we've landed this and pushed it out.

willkg:
That fixes the test that failed and adds a basic test for the celery task and a really basic test for the `sentry_test.py` script.

We could do a lot more to test these things, but this is a cli we're using to set up a new environment and I don't think we should put in a lot of effort to add more maintenance burden over time for something we're likely to use once or twice.

ichnaea/data/tasks.py
```diff
@@ -82,6 +82,11 @@ def monitor_queue_size(self):
     monitor.QueueSize(self)()
 
 
+@celery_app.task(base=BaseTask, bind=True, queue="celery_monitor")
+def sentry_test(self, msg):
+    monitor.SentryTest(self, msg=msg)()
```
jwhitlock: This could be a regular function call (``self.raven_client.captureMessage(msg)``) rather than a callable class instance. It would break the pattern of the other tasks, but I think that would be a good thing.

ichnaea/data/tests/test_monitor.py
```diff
@@ -5,11 +5,12 @@
     monitor_api_key_limits,
     monitor_api_users,
     monitor_queue_size,
+    sentry_test,
 )
 from ichnaea import util
 
 
-class TestMonitorApiKeys(object):
+class TestMonitorApiKeys:
```
jwhitlock: 👍 to Python 3 class definitions

ichnaea/data/monitor.py
```diff
@@ -79,3 +79,14 @@ def __call__(self):
         for name in export_queues | self.task.app.all_queues:
             value = self.task.redis_client.llen(name)
             METRICS.gauge("queue", value, tags=["queue:" + name])
+
+
+class SentryTest(object):
```
jwhitlock: 👎 to Python 2 class definitions

ichnaea/log.py
```diff
@@ -67,8 +67,11 @@ def configure_logging():
 
 
 RAVEN_TRANSPORTS = {
+    # Used in the webapp
     "gevent": GeventedHTTPTransport,
+    # Used in the tests
     "sync": HTTPTransport,
+    # Used in by celery
```
jwhitlock: "Used in by celery" → "Used by celery"

ichnaea/data/monitor.py
willkg: Oo--I missed these. I'll take them out.

ichnaea/data/tasks.py
willkg: I like this idea. +1

willkg:
^^^ That fixes issues raised in comments. Once the tests pass, I'll merge that.

PR #974, opened 2019-10-25
Add a LOGGING_LEVEL setting

### LINKED ISSUE #966: add a debugmode
@ckolos mentioned it'd be handy to have an environment-variable-enabled debug mode for the webapp, celery scheduler, and celery worker. This would help debug infrastructure and startup issues.

When I redid configuration, I changed it so that the logging level is no longer configurable.

This covers figuring something out here.

willkg:
This adds a `LOGGING_LEVEL` setting to make it easier to debug things.

Fixes #966

### DIFF
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -3,7 +3,7 @@
 # Whether or not this is a local development environment.
 # NOTE(willkg): Do not set this in server environments!
 LOCAL_DEV_ENV=True
-LOGGING_LEVE=INFO
+LOGGING_LEVEL=INFO
 
 # Redis
 REDIS_URI=redis://redis:6379/0
--- a/docker/config/local_dev.env
+++ b/docker/config/local_dev.env
@@ -3,6 +3,7 @@
 # Whether or not this is a local development environment.
 # NOTE(willkg): Do not set this in server environments!
 LOCAL_DEV_ENV=True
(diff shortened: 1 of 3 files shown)


ichnaea/conf.py
```diff
@@ -30,6 +38,14 @@ class AppConfig(RequiredConfigMixin):
         parser=bool,
         doc="Whether or not we are running tests.",
     )
+    required_config.add_option(
+        "logging_level",
+        default="INFO",
```
willkg: I think we want INFO in stage/production. I think we want to keep this value and then go through and fix the logging so that we get what we want if stage/production are at INFO level.

docker/config/local_dev.env
```diff
@@ -3,6 +3,7 @@
 # Whether or not this is a local development environment.
 # NOTE(willkg): Do not set this in server environments!
 LOCAL_DEV_ENV=True
+LOGGING_LEVE=INFO
```
jwhitlock: ``LOGGING_LEVE`` → ``LOGGING_LEVEL``

willkg: The next version of Everett will do what-you-meant configuration. I bet we can throw some AI and ML at it and it'll be epic.

willkg:
Thank you!

PR #978, opened 2019-10-29
Update black to 19.10b0

### LINKED ISSUE #977: update black to 19.10b0
Black 19.10b0 came out. We're a couple of days from "first of the month", but since black can require additional changes because it formats Python code, it's probably easier to update on its own.

This covers updating.

willkg:
Fixes #977


willkg:
This just updated dependencies that only affect python formatting. self-r+

PR #981, opened 2019-10-31
Remove "tag" from /__version__ data

### LINKED ISSUE #980: /__version__ should not have "tag" key
https://location.stage.mozaws.net/__version__ shows this:

```
{
"commit": "c296daf883a2aae8cd087f34ad3a376dd21bb799",
"version": "",
"source": "https://github.com/mozilla/ichnaea",
"build": "https://circleci.com/gh/mozilla/ichnaea/1838",
"tag": null
}
```

The dockerflow spec doesn't have a "tag" key. We should remove that.

willkg:
Dockerflow doesn't have a "tag" key, so there shouldn't be one in `/__version__`.

Fixes #980

### DIFF
--- a/ichnaea/util.py
+++ b/ichnaea/util.py
@@ -82,10 +82,6 @@ def version_info():
                 info = json.load(fp)
             except json.JsonDecodeException:
                 pass
-
-    info["build"] = info.get("build", None)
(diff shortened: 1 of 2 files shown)


willkg:
This is what we do in socorro. self-r+

PR #994, opened 2019-11-14
Add "pip check" to build step

willkg:
This adds "pip check" to the build step to make sure we've got versions
of Python requirements that work with one another.

It also upgrades kombu to a version that doesn't fail the check.


willkg:
Thank you!

PR #995, opened 2019-11-14
Document the station updating process

jwhitlock:
The station updating process is mentioned in dataflows, but isn't otherwise documented. This adds the details to the algorithms section. I could also see it in the development / deployment section.

This documents code from these source files:

* [ichnaea/data/station.py](https://github.com/mozilla/ichnaea/blob/master/ichnaea/data/station.py) - The flowchart is derived from ``StationState.transition``. ``StationUpdater.update_shard`` has most of the logic for setup, looping, writing data, and emitting metrics.
* [ichnaea/models/observation.py](https://github.com/mozilla/ichnaea/blob/master/ichnaea/models/observation.py) - ``Report.base_weight`` has the calculations for the qualifying metrics, and ``CellObservation.weight``, and other type-specific ``*Observation`` classes, have the signal weight.
* [ichnaea/models/station.py](https://github.com/mozilla/ichnaea/blob/master/ichnaea/models/station.py) - ``station_blocked`` determines if a station remains blocked.

I don't think code references make sense in the documentation, but if you want the supporting code for a documentation section, I can add it to this PR.


jwhitlock:
Thank you!

hannosch:
Nice work! I’m glad you could figure this all out from the code.

A note for the weights: These are based on my “expert guesses”. I’ve designed these curves using a bunch of example data and looking at and tweaking graphs in a spreadsheet application until they “looked good”. Since we never had “ground truth” data, there wasn’t a good way to validate or improve the weights. Ideally you’d want some kind of learning algorithm for this.

One missing weight is the device type and software version. Some chipsets and operating system versions have known bugs, misreporting signal strength or doing extremely long caching of discovered networks. Some devices also have much better antennas than others. We could never figure out the API for reporting those, while not making the queries too personally identifiable.

Another thing to mention is what the station position in the database actually represents. The algorithm isn’t really trying to figure out, where the wifi access point or cell tower really is, but where users usually are, when they see these signals. A typical situation is someone walking or driving down a street. They’ll see the wifi signals from the buildings on either side of the street. Since they are under an open sky, they’ll probably also get a GNSS signal and might submit data to us. We end up estimating the wifi position somewhere on the street. This is usually also where most people will be, when they see the wifi network. We rarely get submissions from people from inside the building and they wouldn’t have a GNSS signal in there, so we’d not consider them or weigh the les.

A similar situation happens with cell networks. Most modern cell networks use directed antennas, so the area isn’t a circle with the tower in the middle, but rather a triangle with the tower in one of the corners. But we end up using an enclosing circle around the triangle with the center of the triangle as the assumed position. Usually users are somewhere in the covered area and not right next to the tower, so this isn’t too bad. We end up minimizing the accuracy/radius of the returned estimate this way, but don’t actually know where the tower is.

This does get a bit weird, when we combine the position estimates from multiple signals and try to come up with a combined answer. Since we don’t know where the actual signal emitters are, the correlation between signal strength and distance gets really weak. This results in fairly conservative and large accuracy results in our responses.

jwhitlock:
Thanks @hannosch, this is useful stuff.

> A note for the weights: These are based on my “expert guesses”. I’ve designed these curves using a bunch of example data and looking at and tweaking graphs in a spreadsheet application until they “looked good”. Since we never had “ground truth” data, there wasn’t a good way to validate or improve the weights. Ideally you’d want some kind of learning algorithm for this.

This confirms my own guesses looking at the code. I'm thinking of ways to acquire ground truth data for a learning algorithm. It seems like there would be a database of cell tower locations, such as an FCC registration in the US. Improving the accuracy of lookups won't be a priority this year.

> Another thing to mention is what the station position in the database actually represents. The algorithm isn’t really trying to figure out, where the wifi access point or cell tower really is, but where users usually are, when they see these signals.

I came to similar conclusions while reading the code and looking at visualizations of observations on https://www.cellmapper.net. Most observations are on roads, which makes sense for when devices are watching for radios and have GPS etc. enabled. After some thought, it also makes sense that this would still be useful for region lookups, and be "good enough" for many location queries.

> A similar situation happens with cell networks. Most modern cell networks use directed antennas, so the area isn’t a circle with the tower in the middle, but rather a triangle with the tower in one of the corners.

I think this deserves a new paragraph and illustration in the docs. I'm planning another "algorithm" doc on location queries, and may add it with that one.

When skimming the location query code, it appears that the "cell area" position is used for estimates. My initial understanding was that this is a fallback, in case you see a new cell tower but know others in the same area. However, the next time I skimmed the code, it looked like the cell area position was preferred over the cell radio positions. I plan a close reading to determine which way the code goes, but if you have information on cell areas vs cell radios, I'd appreciate it.

hannosch:
> When skimming the location query code, it appears that the "cell area" position is used for estimates. My initial understanding was that this is a fallback, in case you see a new cell tower but know others in the same area. However, the next time I skimmed the code, it looked like the cell area position was preferred over the cell radio positions. I plan a close reading to determine which way the code goes, but if you have information on cell areas vs cell radios, I'd appreciate it.

The code should prefer an individual cell over an area in most cases.

There are three steps to this:

1. In https://github.com/mozilla/ichnaea/blob/master/ichnaea/api/locate/score.py#L54 you have the two sample scoring functions for cells and areas. For a cell, the sample score is based on the number of samples/observations. This is likely going to be in the hundreds or thousands or more. For areas it is not based on the sum of the observations for all cells in the area. Instead it uses the number of cells in the area as the sample count. This is likely going to be a much lower number, maybe we've seen 5 or 10 cells for an area. So ideally the resulting score should be lower. If it is a larger value, this means we have a whole bunch of cells for the area and so the area likely covers a huge region.

2. In https://github.com/mozilla/ichnaea/blob/master/ichnaea/api/locate/cell.py#L311 ``CellPositionMixin.search_cell`` the found cells and areas are clustered and some values are calculated before they are added as a ``PositionResult`` into the result list. At this point some lower and upper bounds are set on the ``accuracy`` value. Looking at those constants in https://github.com/mozilla/ichnaea/blob/master/ichnaea/api/locate/constants.py#L48 you'll notice that an area has a minimum accuracy of 50000 meters and a cell has the same value as an upper bound. So the accuracy value of an area is always going to be larger than that of a cell. If we have an area with a high score, the accuracy is probably also going to be even higher and possibly covers hundreds of kilometers.

3. In https://github.com/mozilla/ichnaea/blob/master/ichnaea/api/locate/result.py#L144 ``PositionResultList`` in the ``best_cluster`` and ``best`` methods the results are sorted by minimum accuracy. A ``PositionResult`` here is an already aggregated result, so it might represent a collection of wifi networks, a collection of cells or a single area. We might have multiple of these for each `type`, so for example three wifi clusters, one cell cluster and two cell areas, where we didn't have any data about the individual cell in the second area.

Arriving at the final answer is a two step process:

First we take all the position results and cluster them together again. Basically each position result represents a circle with the lat/lon as the center point and the accuracy value representing the radius of the circle. We allow for some overlap and add a 50% buffer. This gives us clusters of `matching` or overlapping circles. Or put otherwise, results which `agree` with each other. We than calculate the cumulative score of each of these agreeing estimates and take the cluster with the highest score.

As a second step the `best` function picks out the best single answer from the cluster. It does this by picking the result with the lowest accuracy or put otherwise the most precise.

At this point the min/max bounds mentioned above kick in again. If we have a single cell and its cell area, the area will always have a larger accuracy/radius than the single cell. So this last step is always going to pick out the individual cell. 

I hope this was somewhat comprehensible :-) This would probably benefit from a couple of images with those circles to illustrate each step, but alas time is limited.

PR #996, opened 2019-11-14
Change "umts" to "wcdma" in models docstring

### LINKED ISSUE #989: Clarify documentation / comments for UMTS stations

jwhitlock:
"umts" is the deprecated name for WCDMA stations. This brings the JSON sample in line with the surrounding documentation.

This addresses some of #989.

### DIFF
--- a/ichnaea/models/cell.py
+++ b/ichnaea/models/cell.py
@@ -74,7 +74,7 @@ Example:
 .. code-block:: javascript
 
     {
-        "radio": "umts",
+        "radio": "wcdma",
         "mcc": 123,
         "mnc": 123,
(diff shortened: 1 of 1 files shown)


jwhitlock:
@willkg thanks for the review. After sleeping on it, there are a couple of related changes that I'd like to make:

* Combine definitions of `wcdma` and `umts` into one line in `ichnaea/models/constants.py`.
* Set the names of links in the [API History docs](https://ichnaea.readthedocs.io/en/latest/api/index.html#history), rather than automatically use the doc titles.

I think I'll do these in a new PR, to avoid complicating this review. But, this will no longer be enough to close the issue.

Issue #990, opened 2019-11-05
StaleDataError: UPDATE statement on table 'wifi_shard_N' expected to update X rows; Y were matched


willkg:
We're testing a new infrastructure, 4 years of changes, and an Python 2 -> 3 migration all on stage right now. One of the things we did was add a geosubmit export to prod which points to stage, so stage has a steady stream of incoming reports to process.

One of the errors we're hitting is this:

```
StaleDataError: UPDATE statement on table 'wifi_shard_8' expected to update 457 row(s); 423 were matched.
  File "ichnaea/taskapp/task.py", line 79, in __call__
    result = super(BaseTask, self).__call__(*args, **kw)
  File "celery/app/trace.py", line 648, in __protected_call__
    return self.run(*args, **kwargs)
  File "ichnaea/data/tasks.py", line 145, in update_wifi
    station.WifiUpdater(self, shard_id=shard_id)()
  File "ichnaea/data/station.py", line 566, in __call__
    session, shard, shard_values, stats_counter
  File "ichnaea/data/station.py", line 530, in update_shard
    session.bulk_update_mappings(shard, updates)
  File "sqlalchemy/orm/session.py", line 2860, in bulk_update_mappings
    mapper, mappings, True, False, False, False, False
  File "sqlalchemy/orm/session.py", line 2899, in _bulk_save_mappings
    transaction.rollback(_capture_exception=True)
  File "sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "sqlalchemy/util/compat.py", line 153, in reraise
    raise value
  File "sqlalchemy/orm/session.py", line 2884, in _bulk_save_mappings
    update_changed_only,
  File "sqlalchemy/orm/persistence.py", line 180, in _bulk_update
    bookkeeping=False,
  File "sqlalchemy/orm/persistence.py", line 1029, in _emit_update_statements
    % (table.description, len(records), rows)
```

https://sentry.prod.mozaws.net/operations/location-stage/issues/6582655/

The code in question is here:

https://github.com/mozilla/ichnaea/blob/4b7aa181ddd708c3f0ab9fa6031c49a707d07d69/ichnaea/data/station.py#L530

We're not seeing these errors in prod. Maybe this is caused by different code, dependencies, or data context between stage and prod?

This issue covers looking into this.

hannosch:
[...]

PR #997, opened 2019-11-21
Use pytest.mark.parametrize instead of for loops

jwhitlock:
Use ``pytest.mark.parametrize`` instead of a for loop for the two main types of observation reports.

This dedents the test code, which can mostly be hidden if whitespace is ignored, but some lines are also reformatted with Black.

There are no test changes except for ``test_new``, which now expects one set of metrics per test run instead of two sets, one for each report type.

I did this in preparation of writing a failing test for issue #990, but it appears that this is a database-level issue, not an application issue, so I wasn't able to write a failing unit test. This still feels like a useful refactor, however.

### DIFF
--- a/ichnaea/data/tests/test_station.py
+++ b/ichnaea/data/tests/test_station.py
@@ -256,38 +256,30 @@ class StationTest(BaseStationTest):
         self.check_statcounter(redis, self.stat_obs_key, 3)
         self.check_statcounter(redis, self.stat_station_key, 2)
 
-    def test_new(self, celery, session, metricsmock):
-        for source in (ReportSource.gnss, ReportSource.query):
-            obs = self.obs_factory.build(source=source)
-            obs1 = self.obs_factory(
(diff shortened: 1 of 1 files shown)


ichnaea/data/tests/test_station.py
```diff
@@ -256,143 +256,135 @@ def test_blocklist_skip(self, celery, redis, session):
...
+        self.check_dates(station, self.today, self.today, self.today)
 
         assert (
             len(
                 metricsmock.filter_records(
                     "incr", "data.observation.insert", value=5, tags=[self.type_tag]
                 )
             )
-            == 2
+            == 1
```
jwhitlock: Now that `queue_and_update` is run once per test rather than twice in a loop, this metric, and the following `data.station.new` metric, is only incremented once.

jwhitlock:
Thanks!

My GitHub diff looked decent in "Split" view with "Hide whitespace changes" selected.

PR #998, opened 2019-11-21
Add make runwebdebug and make redis

jwhitlock:
These were helpful when I was figuring out the station updating data pipeline, for testing staging and writing the observations doc. They may be more generally useful, or maybe they should just be sections in the developer documentation.

* ``make runwebdebug`` - Run the webapp in a single thread outside gunicorn. This allows you to add pdb breakpoints and walk through the code.
* ``make redis`` - Open a Redis CLI session

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,12 +29,14 @@ default:
 	@echo "  build            - build docker containers"
 	@echo "  setup            - drop and recreate service state"
 	@echo "  run              - run webapp"
+	@echo "  runwebdebug      - run webapp outside gunicorn, for debugging"
 	@echo "  runcelery        - run scheduler and worker"
 	@echo "  runservices      - run service containers (mysql, redis, etc)"
 	@echo "  stop             - stop all service containers"
 	@echo ""
 	@echo "  shell            - open a shell in the app container"
 	@echo "  mysql            - open mysql prompt"
+	@echo "  redis            - open redis prompt"
 	@echo "  clean            - remove all build, test, coverage and Python artifacts"
 	@echo "  lint             - lint code"
 	@echo "  lintfix          - reformat code"
@@ -84,6 +86,11 @@ mysql: my.env .docker-build
 	${DC} up -d mysql
 	${DC} exec mysql mysql --user root --password=location location
 
+.PHONY: redis
+redis: my.env .docker-build
+	${DC} up -d redis
+	${DC} exec redis redis-cli
+
 .PHONY: test
 test: my.env .docker-build
 	./bin/test_env.sh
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -120,6 +127,10 @@ lintfix: my.env .docker-build
 run: my.env .docker-build
 	${DC} up web
 
+.PHONY: runwebdebug
+runwebdebug: my.env .docker-build
+	${DC} run --rm --service-ports web shell python ichnaea/webapp/app.py
```
willkg: Doesn't `make run` already do this? It runs `web` which runs `docker/run_web.sh` which sees `LOCAL_DEV_ENV=true` and runs the app the way you're suggesting here.

Makefile
```diff
@@ -29,12 +29,14 @@ default:
 	@echo "  build            - build docker containers"
 	@echo "  setup            - drop and recreate service state"
 	@echo "  run              - run webapp"
+	@echo "  runwebdebug      - run webapp outside gunicorn, for debugging"
```
jwhitlock: This comment is wrong. ``make run`` runs the app inside of ``waitress``, not ``gunicorn``.

Makefile
jwhitlock: The difference is when I add a ``pdb`` breakpoint. For example, when I add ``import pdb; pdb.set_trace()`` to ``ichnaea/content/views.py``, ``homepage_view``, and refresh http://localhost:8000, I get a 500 Internal Server Error with ``make run``, and an interactive ``pdb`` session with ``make runwebdebug``.

Also, it seems useful to shutdown the ``web`` container when running this, otherwise it needs to be done manually.

willkg: I think I'd rather have a `webshell` make rule like this:

```
.PHONY: webshell
webshell: my.env .docker-build
    ${DC} run --rm --service-ports web shell
```

That gets you into a web context and covers running the web server with pdb and some other things.

Thinking ahead, I want to redo how static assets are built, so then the `node.make` thing will go away. This will be useful for that, too.

jwhitlock:
I think I'm going to move `make runwebdebug` to my personal "snippets". It's not that much harder to paste into my shell:

```
docker-compose stop web
docker-compose run --rm --service-ports web shell python ichnaea/webapp/app.py
```

and then I don't have to document why you'd want to use it.

Speaking of documentation, ``make redis`` is already basically documented, but manually:

https://ichnaea.readthedocs.io/en/latest/debug.html#connections

I'd like to wait to make these changes until we increase from N=1 to N=2 for usefulness, and get the developer docs in sync with any features we build.

PR #999, opened 2019-11-21
Make it clearer that wcdma is alias for umts

### LINKED ISSUE #989: Clarify documentation / comments for UMTS stations

jwhitlock:
``wcdma`` is the preferred name for ``umts``. This is the remaining task to fix #989.

I also considered removing instance of ``Radio.umts`` from the codebase. This would be less than an hour of work, and could be done instead of this change.


jwhitlock:
I'd rather remove `umts`, reworking this PR.

PR #1000, opened 2019-11-21
Update API History section

jwhitlock:
The document titles, like "Geosubmit: /v1/geosubmit (DEPRECATED)", no longer work as the link text, making the [History section](https://ichnaea.readthedocs.io/en/latest/api/index.html#history) nonsensical: 

> The service first offered the custom /v1/search/ [Submit: /v1/submit (DEPRECATED)](https://ichnaea.readthedocs.io/en/latest/api/submit.html#api-submit) APIs. Later it was decided to also implement the [Geolocate: /v1/geolocate](https://ichnaea.readthedocs.io/en/latest/api/geolocate.html#api-geolocate) API to lessen the burden on clients that want to support multiple location services. As an extension to this the [Geosubmit: /v1/geosubmit (DEPRECATED)](https://ichnaea.readthedocs.io/en/latest/api/geosubmit.html#api-geosubmit) API was added to offer a consistent way to contribute back data to the service. Afterwards the [Region: /v1/country](https://ichnaea.readthedocs.io/en/latest/api/region.html#api-region) API was added and [Geosubmit Version 2: /v2/geosubmit](https://ichnaea.readthedocs.io/en/latest/api/geosubmit2.html#api-geosubmit2) superseded its version 1 counterpart.

Instead, use the API endpoint path as the link text. Organize history by years, and fill in some missing API history.

My research suggests that the `/v1/submit` API can also be removed, as it was removed from the MozStumbler code in 2016.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -159,9 +159,34 @@ APIs
 History
 =======
 
-The service first offered the custom ``/v1/search/`` :ref:`api_submit` APIs.
-Later it was decided to also implement the :ref:`api_geolocate` API to lessen
-the burden on clients that want to support multiple location services. As an
-extension to this the :ref:`api_geosubmit` API was added to offer a consistent
-way to contribute back data to the service. Afterwards the :ref:`api_region`
-API was added and :ref:`api_geosubmit2` superseded its version 1 counterpart.
+The service launched in 2013, and first offered custom ``/v1/search/`` and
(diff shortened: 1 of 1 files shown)


hannosch:
> My research suggests that the `/v1/submit` API can also be removed, as it was removed from the MozStumbler code in 2016

The v1/submit endpoint should still be used by the stumbler code inside Fennec. Which is essentially a very old fork of the MozStumbler code. But it is also the source of most of the data contributions.

PR #1001, opened 2019-11-22
Remove Radio.umts

### LINKED ISSUE #989: Clarify documentation / comments for UMTS stations

jwhitlock:
``Radio.umts`` was the deprecated alias for ``Radio.wcdma``. The deprecated APIs still refer to this station type as "umts", and ``UMTS`` is used in the public export, but otherwise the code refers to it as ``wcdma`` now.

The largest changes are to the public export tests, many of which I wrote for the new ``load_cell_data.py`` feature. This may explain why I'm personally invested in removing as much "umts" as I can. 😁

This should fix #989. It replaces a smaller change in #999.


jwhitlock:
Thanks!

PR #1003, opened 2019-11-26
bug 1598325: fix None comparison error

willkg:
This catches one I missed in the previous pass.


willkg:
This is just like pr #1002. If that one is bad, then this one is, too. self-r+

PR #1004, opened 2019-11-27
976: Add release script and configuration

### LINKED ISSUE #976: add release.py script
For Tecken, Socorro, and Antenna, we have a release script which is maintained here:

https://github.com/willkg/socorro-release/

We should use that here, too.

This issue covers adding the script and the required configuration file.

willkg:
This adds a release script and corresponding configuration so we can automate releases more.

Fixes #976

### DIFF
--- a/bin/release.py
+++ b/bin/release.py
@@ -1,0 +1,320 @@
+#!/usr/bin/env python
+
+# This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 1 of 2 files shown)


willkg:
Unfortunately, this is hard to test without being in the master branch. We use the same script in Socorro, Antenna, and Tecken, so it's probably fine.

willkg:
Thank you!

PR #1006, opened 2019-12-03
bug 1600706: change update_incoming batch size from 100 to 500

willkg:
This will cause a single celery task to process 500 rather than 100. Overall, this will reduce the amount of bookkeeping done per `update_incoming` item. The mean time for executing an `update_incoming`
task is pretty small compared to the other tasks, so this seems like a safe try.


willkg:
I was thinking the same thing, but I figured 100 -> 500 should be safe and noticeable in the dashboard if it helped and if it creates other problems. And hopefully it gives us a good feeling for a better number.

willkg:
Thank you!

PR #1007, opened 2019-12-04
bug 1600706: increase update_incoming batch size to 5000

willkg:
We changed the batch size from 100 to 500 yesterday and that didn't affect the graphs in a noticeable way. It also didn't seem to affect other things, so I think it's safe to increase it more.


willkg:
Thank you!

PR #1008, opened 2019-12-05
Replace sync_session fixture with session

jwhitlock:
After removing the alternate database drivers, the two session fixtures are identical, and tests using the ``sync_session`` fixture work just as well with the more common ``session`` fixture.

### DIFF
--- a/ichnaea/conftest.py
+++ b/ichnaea/conftest.py
@@ -196,25 +196,6 @@ def session(db):
 
 
 @pytest.fixture(scope="function")
-def sync_session(db):
-    with db.engine.connect() as conn:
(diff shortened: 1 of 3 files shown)


willkg:
I have vague memories of them being different, so it's possible I either messed that up or they were different and then I made changes such that they weren't.

I'll double-check what I did just to make sure I didn't mess something else up, then review this.

willkg:
Looks like I made the two fixtures identical. I didn't find other notes about why I did what I did or the progression. It probably means it was an oversight to keep the `sync_session` fixture and that this is a good change.

jwhitlock:
Thanks!

PR #1009, opened 2019-12-05
Fix StationUpdate deadlock handler and track handled deadlocks

### LINKED ISSUE #991: OperationalError: pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction' ...
We're testing a new infrastructure, 4 years of changes, and an Python 2 -> 3 migration all on stage right now. One of the things we did was add a geosubmit export to prod which points to stage, so stage has a steady stream of incoming reports to process.

One of the errors we're hitting is this:

```
OperationalError: (1213, 'Deadlock found when trying to get lock; try restarting transaction')
  File "sqlalchemy/engine/base.py", line 1229, in _execute_context
    cursor, statement, parameters, context
  File "sqlalchemy/dialects/mysql/mysqldb.py", line 148, in do_executemany
    rowcount = cursor.executemany(statement, parameters)
  File "pymysql/cursors.py", line 199, in executemany
    self.rowcount = sum(self.execute(query, arg) for arg in args)
  File "pymysql/cursors.py", line 199, in <genexpr>
    self.rowcount = sum(self.execute(query, arg) for arg in args)
  File "pymysql/cursors.py", line 170, in execute
    result = self._query(query)
  File "pymysql/cursors.py", line 328, in _query
    conn.query(q)
  File "pymysql/connections.py", line 517, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "pymysql/connections.py", line 732, in _read_query_result
    result.read()
  File "pymysql/connections.py", line 1075, in read
    first_packet = self.connection._read_packet()
  File "pymysql/connections.py", line 684, in _read_packet
    packet.check_error()
  File "pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)

OperationalError: (pymysql.err.OperationalError) (1213, 'Deadlock found when trying to get lock; try restarting transaction')
[SQL: UPDATE wifi_shard_8 SET max_lat=%(max_lat)s, min_lat=%(min_lat)s, max_lon=%(max_lon)s, min_lon=%(min_lon)s, lat=%(lat)s, lon=%(lon)s, modified=%(modified)s, radius=%(radius)s, region=%(region)s, samples=%(samples)s, source=%(source)s, weight=%(weight)s, last_seen=%(last_seen)s, block_first=%(block_first)s, block_last=%(block_last)s, block_count=%(block_count)s WHERE wifi_shard_8.mac = %(wifi_shard_8_mac)s]
[parameters: (REDACTED)}  ... displaying 10 of 432 tota...
  File "ichnaea/taskapp/task.py", line 79, in __call__
    result = super(BaseTask, self).__call__(*args, **kw)
  File "celery/app/trace.py", line 648, in __protected_call__
    return self.run(*args, **kwargs)
  File "ichnaea/data/tasks.py", line 145, in update_wifi
    station.WifiUpdater(self, shard_id=shard_id)()
  File "ichnaea/data/station.py", line 566, in __call__
    session, shard, shard_values, stats_counter
  File "ichnaea/data/station.py", line 530, in update_shard
    session.bulk_update_mappings(shard, updates)
  File "sqlalchemy/orm/session.py", line 2860, in bulk_update_mappings
    mapper, mappings, True, False, False, False, False
  File "sqlalchemy/orm/session.py", line 2899, in _bulk_save_mappings
    transaction.rollback(_capture_exception=True)
  File "sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "sqlalchemy/util/compat.py", line 153, in reraise
    raise value
  File "sqlalchemy/orm/session.py", line 2884, in _bulk_save_mappings
    update_changed_only,
  File "sqlalchemy/orm/persistence.py", line 180, in _bulk_update
    bookkeeping=False,
  File "sqlalchemy/orm/persistence.py", line 996, in _emit_update_statements
    statement, multiparams
  File "sqlalchemy/engine/base.py", line 988, in execute
    return meth(self, multiparams, params)
  File "sqlalchemy/sql/elements.py", line 287, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "sqlalchemy/engine/base.py", line 1107, in _execute_clauseelement
    distilled_params,
  File "sqlalchemy/engine/base.py", line 1253, in _execute_context
    e, statement, parameters, cursor, context
  File "sqlalchemy/engine/base.py", line 1473, in _handle_dbapi_exception
    util.raise_from_cause(sqlalchemy_exception, exc_info)
  File "sqlalchemy/util/compat.py", line 398, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "sqlalchemy/util/compat.py", line 152, in reraise
    raise value.with_traceback(tb)
  File "sqlalchemy/engine/base.py", line 1229, in _execute_context
    cursor, statement, parameters, context
  File "sqlalchemy/dialects/mysql/mysqldb.py", line 148, in do_executemany
    rowcount = cursor.executemany(statement, parameters)
  File "pymysql/cursors.py", line 199, in executemany
    self.rowcount = sum(self.execute(query, arg) for arg in args)
  File "pymysql/cursors.py", line 199, in <genexpr>
    self.rowcount = sum(self.execute(query, arg) for arg in args)
  File "pymysql/cursors.py", line 170, in execute
    result = self._query(query)
  File "pymysql/cursors.py", line 328, in _query
    conn.query(q)
  File "pymysql/connections.py", line 517, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "pymysql/connections.py", line 732, in _read_query_result
    result.read()
  File "pymysql/connections.py", line 1075, in read
    first_packet = self.connection._read_packet()
  File "pymysql/connections.py", line 684, in _read_packet
    packet.check_error()
  File "pymysql/protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "pymysql/err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
```

https://sentry.prod.mozaws.net/operations/location-stage/issues/6582659/

The issue is coming from this line:

https://github.com/mozilla/ichnaea/blob/4b7aa181ddd708c3f0ab9fa6031c49a707d07d69/ichnaea/data/station.py#L530

It's happening between 25-30 times per hour.

We're not seeing these errors in prod. Maybe this is caused by different code, dependencies, or data context between stage and prod?

This issue covers looking into this.

This might be related to #990.

jwhitlock:
In [PyMySQL 0.9.0](https://github.com/PyMySQL/PyMySQL/blob/master/CHANGELOG#L39), ``LOCK_DEADLOCK`` changed from an internal error to an operational error. This also changes the SQLAlchemy error class used to wrap it. This broke the deadlock handler, and we started to get many more of these when the latest code went to production.

This PR updates the handler to only look at the MySQL error code, and be more flexible about the error class used to report it. It also increments a metric ``data.station.dberror``, to allow us to track how handled deadlocks are affected by other data pipeline changes. I also added some tests to cover this behaviour, that may have caught this change.

I struggled with some weirdness around testing code that sends a ``ROLLBACK``, such as this code when detecting any database error. Because the same database connection is used in tests and in the code being tested, the ``ROLLBACK`` was causing test rows to be rolled back as well, changing the database for the next round. For an existing test, this changed an "UPDATE" operation into an "INSERT" operation, when the ``ROLLBACK`` removed the cell from the database.

I've fixed this by using ``session.begin_nested()`` in the task session, to create a nested transaction. I tried some other things, such as creating a new database connection, and registering an event handler as suggested by the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html?highlight=test#joining-a-session-into-an-external-transaction-such-as-for-test-suites), but this was the most reliable method. With the way it is written, it will also use nested transactions in production, which should be safe but might have other effects. If we're worried about that, I can try to restrict the nested transaction to tests.

I also added a second mode to the `raven` test fixture. Previously, it asserted that no messages were sent with raven during the test in the fixture tear down. That is still the default, but now setting `raven.expect_message = True` will change to asserting that at least one message was sent.

This should fix #991, but I believe there are more changes to avoid deadlocks in the first place.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -33,6 +33,7 @@ Metric Name                      Type    Tags
 `data.report.upload`_            counter key
 `data.station.blocklist`_        counter type
 `data.station.confirm`_          counter type
+`data.station.dberror`_          counter type, errno
 `data.station.new`_              counter type
 `datamaps`_                      timer   func, count
 `locate.fallback.cache`_         counter fallback_name, status
@@ -285,6 +286,7 @@ The fallback name tag specifies which fallback service is used.
 .. _data.station.confirm:
 .. _data.station.blocklist:
 .. _data.station.new:
+.. _data.station.dberror:
 
 Data Pipeline Metrics
 ---------------------
@@ -381,6 +383,18 @@ Along the way several counters measure the steps involved:
 
     Count the number of Bluetooth, cell or WiFi :term:`station` that were
     discovered for the first time.
+
+``data.station.dberror#type:<type>,errno:<errno>``: counters
+
+    Count the number of retryable database errors.  ``type`` is ``blue``,
+    ``cell``, or ``wifi``, and ``errno`` is the error number, which can be
+    found on the `MySQL Server Error Reference`_.
+
+    Retryable database errors, like a deadlock (``1213``) cause the station
+    updating task to sleep and start over. Other database errors are not
+    counted, but instead halt the task and are recorded in Sentry.
+
+.. _`MySQL Server Error Reference`: https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html
 
 .. _data.export.batch:
 .. _data.export.upload:
--- a/ichnaea/data/tests/test_station.py
+++ b/ichnaea/data/tests/test_station.py
@@ -151,6 +151,12 @@ class TestDatabaseErrors(BaseStationTest):
                 )
                 == 1
             )
+            assert metricsmock.has_record(
+                "incr",
+                "data.station.dberror",
+                value=1,
+                tags=["type:cell", "errno:%s" % LOCK_WAIT_TIMEOUT],
+            )
         finally:
             CellUpdater._retry_wait = orig_wait
             session.execute(text("drop table %s;" % cell.__tablename__))
@@ -221,6 +227,12 @@ class TestDatabaseErrors(BaseStationTest):
             "incr", "data.station.confirm", value=1, tags=["type:cell"]
         )
         assert metricsmock.has_record("timing", "task", tags=["task:data.update_cell"])
+        assert metricsmock.has_record(
+            "incr",
+            "data.station.dberror",
+            value=1,
+            tags=["type:cell", "errno:%s" % errno],
+        )
 
 
 class StationTest(BaseStationTest):
--- a/ichnaea/data/tests/test_station.py
+++ b/ichnaea/data/tests/test_station.py
@@ -97,6 +97,8 @@ class TestDatabaseErrors(BaseStationTest):
         )
         session.add(cell)
         session.commit()
+        # TODO: Find a more elegant way to do this
+        db.tests_task_use_savepoint = True
 
         error = errclass(errno, errmsg)
         wrapped = InterfaceError.instance(
@@ -111,6 +113,8 @@ class TestDatabaseErrors(BaseStationTest):
             self._queue_and_update(celery, [obs], update_cell)
             assert CellUpdater.add_area_update.call_count == 2
             sleepy.assert_called_once_with(1)
+
+        del db.tests_task_use_savepoint
 
         cells = session.query(shard).all()
         assert len(cells) == 1
--- a/ichnaea/data/tests/test_station.py
+++ b/ichnaea/data/tests/test_station.py
@@ -6,11 +6,9 @@ import pytest
 from pymysql.err import InternalError, MySQLError, OperationalError
 from pymysql.constants.ER import LOCK_DEADLOCK, LOCK_WAIT_TIMEOUT
 from pytz import UTC
-from sqlalchemy import text
 from sqlalchemy.exc import InterfaceError
 
 from geocalc import destination
-from ichnaea.db import configure_db
 from ichnaea.data.station import CellUpdater
 from ichnaea.data.tasks import update_blue, update_cell, update_wifi
 from ichnaea.models import (
@@ -61,105 +59,9 @@ class BaseStationTest:
 
 
 class TestDatabaseErrors(BaseStationTest):
-    # this is a standalone class to ensure DB isolation
-
     queue_prefix = "update_cell_"
     shard_model = CellShard
     unique_key = "cellid"
-
-    @pytest.fixture(scope="function")
-    def session2(self, database):
-        # Create a second independent database session.
-        db = configure_db("ro")
-        session = db.session()
-        try:
-            yield session
-            session.rollback()
-        finally:
-            session.close()
-            db.close()
-
-    def queue_and_update(self, celery, obs):
-        return self._queue_and_update(celery, obs, update_cell)
-
-    def test_lock_timeout(
-        self, celery, redis, session, session2, metricsmock, restore_db
-    ):
-        obs = CellObservationFactory.build()
-        cell = CellShardFactory.build(
-            radio=obs.radio,
-            mcc=obs.mcc,
-            mnc=obs.mnc,
-            lac=obs.lac,
-            cid=obs.cid,
-            samples=10,
-            created=datetime(2019, 12, 1, tzinfo=UTC),
-        )
-        session2.add(cell)
-        # flush sends the INSERT to the database, but does not commit it.
-        # This cell is seen as an existing cell the first pass through
-        # the task. However, because of the rollback inside the task, it is
-        # removed from the database, and the observation inserts it as a new
-        # cell. This is probably an error in the test, but it retained because
-        # it is a working historical test.
-        session2.flush()
-
-        orig_add_area = CellUpdater.add_area_update
-        orig_wait = CellUpdater._retry_wait
-        num = [0]
-
-        def mock_area(self, updated_areas, key, num=num, session2=session2):
-            orig_add_area(self, updated_areas, key)
-            num[0] += 1
-            if num[0] == 2:
-                session2.rollback()
-
-        try:
-            CellUpdater._retry_wait = 0.0001
-            session.execute("set session innodb_lock_wait_timeout = 1")
-            with mock.patch.object(CellUpdater, "add_area_update", mock_area):
-                self.queue_and_update(celery, [obs])
-
-            # the inner task logic was called exactly twice
-            assert num[0] == 2
-
-            shard = CellShard.shard_model(obs.cellid)
-            cells = session.query(shard).all()
-            assert len(cells) == 1
-            self.check_statcounter(redis, StatKey.cell, 1)
-
-            # As mentioned above, the cell was inserted as new from the observation
-            cell = cells[0]
-            assert cell.samples == 1
-            assert datetime.now(tz=UTC) - cell.created < timedelta(seconds=60)
-            self.check_statcounter(redis, StatKey.unique_cell, 1)
-
-            # Assert generated metrics are correct
-            assert (
-                len(
-                    metricsmock.filter_records(
-                        "incr", "data.observation.insert", value=1, tags=["type:cell"]
-                    )
-                )
-                == 1
-            )
-            assert (
-                len(
-                    metricsmock.filter_records(
-                        "timing", "task", tags=["task:data.update_cell"]
-                    )
-                )
-                == 1
-            )
-            assert metricsmock.has_record(
-                "incr",
-                "data.station.dberror",
-                value=1,
-                tags=["type:cell", "errno:%s" % LOCK_WAIT_TIMEOUT],
-            )
-        finally:
-            CellUpdater._retry_wait = orig_wait
-            session.execute(text("drop table %s;" % cell.__tablename__))
 
     @pytest.mark.parametrize(
         "errclass,errno,errmsg",
@@ -180,6 +82,7 @@ class TestDatabaseErrors(BaseStationTest):
     def test_retriable_exceptions(
         self, celery, redis, session, db, metricsmock, errclass, errno, errmsg
     ):
+        """Test database exceptions where the task should wait and try again."""
 
         obs = CellObservationFactory.build(radio=Radio.lte)
         shard = CellShard.shard_model(obs.cellid)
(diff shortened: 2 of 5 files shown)


docs/metrics.rst
```diff
@@ -382,6 +384,18 @@ Along the way several counters measure the steps involved:
...
+
+    Count the number of retryable database errors.  Type is ``blue``, ``cell``,
+    or ``wifi``, and ``errno`` is the error number, which can be found on the
+    `MySQL Server Error Reference`_.
+
+    Retryable database errors, like a deadlock (``1213``) cause the station
+    updating task to sleep and start over. Other database errors are not
+    counted, but instead halt the task and are recorded in Sentry.
+
+.. _`MySQL Server Error Reference`: https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html
```
willkg: Very helpful! :+1:

ichnaea/data/tests/test_station.py
```diff
@@ -137,10 +151,150 @@ def mock_area(self, updated_areas, key, num=num, session2=session2):
...
 
+    @pytest.mark.parametrize(
+        "errclass,errno,errmsg",
+        (
+            (
+                # This was an InternalError before PyMySQL 0.9.0
+                InternalError,
+                LOCK_DEADLOCK,
+                "Deadlock found when trying to get lock; try restarting transaction",
+            ),
```
willkg: We don't need to support PyMySQL versions we're not using. Why is this test helpful to have going forward?

If we're concerned about LOCK_DEADLOCK being switched to something different, I feel like we should do a different kind of test here, but I can't think of a way to do that without it being really ornery which is its own kind of maintenance struggle.

ichnaea/data/tests/test_station.py
```diff
@@ -137,10 +151,150 @@ def mock_area(self, updated_areas, key, num=num, session2=session2):
...
+        )
+
+    @pytest.mark.parametrize(
+        "errclass,errno,errmsg",
+        (
+            (
+                OperationalError,
+                DBACCESS_DENIED_ERROR,
+                "Access denied for user 'ichnaea'@'localhost' to database 'location'",
+            ),
```
willkg: I don't understand what this test is here for. Is this error different than the others? Do we really need to be testing a situation that's a infrastructure/configuration problem?

ichnaea/data/tests/test_station.py
jwhitlock: That's a good point, I'll drop the legacy test case.

ichnaea/data/tests/test_station.py
jwhitlock: The test is that a non-retryable database error aborts the task. It might show up in a code coverage report. If that's not valuable enough, we can lose the 50 lines.

ichnaea/data/tests/test_station.py
```diff
@@ -89,8 +93,15 @@ def test_lock_timeout(
...
             samples=10,
+            created=datetime(2019, 12, 1, tzinfo=UTC),
         )
         session2.add(cell)
+        # flush sends the INSERT to the database, but does not commit it.
+        # This cell is seen as an existing cell the first pass through
+        # the task. However, because of the rollback inside the task, it is
+        # removed from the database, and the observation inserts it as a new
+        # cell. This is probably an error in the test, but it retained because
+        # it is a working historical test.
```
jwhitlock: I'd support removing this test entirely. It covers the same issue as ``test_retriable_exceptions``, and generates the error in a more "natural" way, but it also relies on the buggy behaviour around session rollbacks.

willkg: If you think the thing we need to test is covered by another test, I'm totally game for removing it.

ichnaea/data/tests/test_station.py
willkg: Oh, I see--you used that error because it wasn't retryable. I think that makes sense to test. I thought you were explicitly trying to test an access error which I didn't think was helpful.

Maybe (in a future PR), we could extract the retry logic and make it a function or a decorator or a context manager or something like we have with Antenna and Socorro. Then we can test the retry functionality separate from whatever needs retrying. I don't know if there are other retryable things in the code base.

Anyhow, having this in or out is fine. Whichever you prefer--your feelings on how the code might evolve over time and what kinds of things will help are insightful here.

jwhitlock: > Maybe (in a future PR), we could extract the retry logic and make it a function or a decorator or a context manager or something like we have with Antenna and Socorro. Then we can test the retry functionality separate from whatever needs retrying. I don't know if there are other retryable things in the code base.

I think only the ``StationUpdater`` has this retry logic, and the Sentry instances of deadlock errors from the 1.5 branch were from other updaters, like the cell area or datamap updaters. So, there may be a chance to make a general way of defining retrying.

There's other things I'd like to try.  Using ``SELECT ... FOR UPDATE`` would move locking earlier in the algorithm, and may eliminate the deadlocks (while increasing lock timeouts). This would make the tradeoffs in chunk size more apparent.

In other projects, I've relied on Celery to rerun the task later on failures. It doesn't work for us because we're pulling data from a Redis queue rather than passing it as arguments. I think this suggests we're using Celery wrong, or maybe Celery is a poor match for this volume of data.

jwhitlock: > Anyhow, having this in or out is fine. Whichever you prefer--your feelings on how the code might evolve over time and what kinds of things will help are insightful here.

I'm inclined to leave it out for now. If we end up needing something like it, we can resurrect this test code.

jwhitlock:
> Getting rid of these deadlocks will be a good thing.

I agree, but I think there were plenty of deadlocks last month with the 1.5.x code, just that there was an error handler so that they were retried. Handling and retrying is the right thing, and should get rid of a lot of Sentry errors. I hope to change the code to reduce deadlocks next, and use the metrics to measure the effects.

ichnaea/db.py
```diff
@@ -83,6 +83,8 @@ def db_worker_session(database, commit=True):
     """
     try:
         session = database.session()
+        if commit:
+            session.begin_nested()
```
jwhitlock: This is the bit that will change task behaviour in production, for more than just station updating. If there is a general slowdown, I'll want to ensure ``begin_nested`` only runs in tests.

hannosch:
For testing session rollbacks, did you see https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites

There’s some setup in the test suite to automatically turn sessions inside tests to sub-transactions. Not sure if that is applicable here.

ichnaea/data/tests/test_station.py
hannosch: > In other projects, I've relied on Celery to rerun the task later on failures. It doesn't work for us because we're pulling data from a Redis queue rather than passing it as arguments. I think this suggests we're using Celery wrong, or maybe Celery is a poor match for this volume of data.

Celery was never really a good fit for this project, but it was simple to get going and ops had experience with it. The data pipeline needs to be able to split/multiplex and change batch sizes of the incoming data (e.g. commit in ~1mb blocks to AWS S3 as each put request costs money and has size limitations, but the same data should be processed faster for the SQL backend). None of those things were available in Celery at the time, so all the custom Redis logic was written. The code was probably influenced by me working on a Kafka based project before. If I had been able to continue work on the project, I’d have probably converted it to use AWS Kinesis for the data pipeline part and AWS Aurora as a MySQL drop-in. In a later step, I would have replaced the database side with a key/value store for the main data and some smaller db for the statistics side.

jwhitlock:
> For testing session rollbacks, did you see https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#joining-a-session-into-an-external-transaction-such-as-for-test-suites

Yes, I saw that article, and tried to use the technique. I was able to attach an "after_transaction_end" event handler to the ``session`` fixture, but the rollback occurs in the dynamically generated ``db_worker_session``, and I couldn't see a way to hook an event onto it in the test environment:

https://github.com/mozilla/ichnaea/blob/8c41057cd41563f26e759c583873489bf7eb5133/ichnaea/db.py#L76-L93

I am interested in exploring this technique to see if we can put "regular" tests in a transaction, and only do the database setup once. Maybe with some refactoring, it can be used in the tasks as well.

hannosch:
> I am interested in exploring this technique to see if we can put "regular" tests in a transaction, and only do the database setup once. Maybe with some refactoring, it can be used in the tasks as well.

The database setup only happens once for all normal tests. That’s what https://github.com/mozilla/ichnaea/blob/8c41057cd41563f26e759c583873489bf7eb5133/ichnaea/conftest.py#L179 should be doing, as long as you use the correct test fixture. The session tracker fixture might be an example on how to attach an event to the current session in the tests and run some code before/after each test function using that fixture.

ichnaea/data/tests/test_station.py
```diff
@@ -141,6 +155,74 @@ def mock_area(self, updated_areas, key, num=num, session2=session2):
...
+    )
+    def test_retriable_exceptions(
+        self, celery, redis, session, db, metricsmock, errclass, errno, errmsg
+    ):
+
+        obs = CellObservationFactory.build(radio=Radio.lte)
+        shard = CellShard.shard_model(obs.cellid)
+        cell = CellShardFactory.build(
+            radio=obs.radio,
+            cc=obs.mcc,
```
jwhitlock: this should be `mcc=, not `cc=`

jwhitlock:
I took another run at figuring out how to ensure the task runs in a nested transaction using just fixtures and events, but it is slow going, and I suspect the changes will be deep. I also think we'll want to figure this out before we roll out retry code for deadlocks across more tasks.

I added more code to further restrict the ``session.begin_nested()`` call, so that it only runs when this test is checking deadlocks and timeouts. That means the production tasks will continue to run the same way, without a ``SAVEPOINT`` call.

This makes this PR safer if a little uglier, and we can see the effects of the working deadlock handler in stage today and over the weekend.

PR #1010, opened 2019-12-06
Restrict SQLAlchemy exception handler

jwhitlock:
SQLAlchemy wraps database exceptions, like a deadlock, in ``StatementError``, and passes the original exception in ``.orig``. Other exceptions, derived from the more generic ``SQLAlchemyError``, are possible, such as a ``StaleDataError``.

This is follow-on work for issue #991 and issue #990.


jwhitlock:
Self r+

willkg:
Nice job! belated r+

PR #1011, opened 2019-12-06
Encode binary IDs in base64 for numpy

### LINKED ISSUE #921: test_locate_v1.test_blue_seen fails occasionally

jwhitlock:
numpy strips trailing ``NULL`` characters (``0x00``) from byte strings ([numpy issue 8089](https://github.com/numpy/numpy/issues/8089)). This can cause stations to be detected as new stations, requiring further processing.

Instead of storing the raw binary IDs in the numpy array, store a base64-encoded string, and decode it before passing back to ``InternalPositionSource``.

Fixes #921, which was randomly hitting this issue with Bluetooth stations.

### DIFF
--- a/ichnaea/api/locate/cell.py
+++ b/ichnaea/api/locate/cell.py
@@ -1,5 +1,6 @@
 """Search implementation using a cell database."""
 
+from base64 import b64decode
 from collections import defaultdict
 import math
 
@@ -42,7 +43,7 @@ NETWORK_DTYPE = numpy.dtype(
         ("age", numpy.int32),
         ("signalStrength", numpy.int32),
         ("score", numpy.double),
-        ("id", "S11"),
+        ("id_b64", "S16"),
         ("seen_today", numpy.bool),
     ]
 )
@@ -79,7 +80,7 @@ def cluster_cells(cells, lookups, min_age=0):
                         obs_data[cell.cellid][0],
                         obs_data[cell.cellid][1],
                         station_score(cell, now),
-                        encode_cellid(*cell.cellid),
+                        encode_cellid(*cell.cellid, codec="base64"),
                         bool(cell.last_seen is not None and cell.last_seen >= today),
                     )
                     for cell in area_cells
@@ -118,7 +119,7 @@ def cluster_areas(areas, lookups, min_age=0):
                         obs_data[area.areaid][0],
                         obs_data[area.areaid][1],
                         area_score(area, now),
-                        encode_cellarea(*area.areaid),
+                        encode_cellarea(*area.areaid, codec="base64"),
                         bool(area.last_seen is not None and area.last_seen >= today),
                     )
                 ],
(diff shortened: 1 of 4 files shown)


ichnaea/api/locate/internal.py
```diff
@@ -56,10 +56,10 @@ def _store_query(self, query, results):
             return
 
         result_networks = {"area": set(), "blue": set(), "cell": set(), "wifi": set()}
-        for network in best_result.used_networks:
-            if network[2]:
+        for net_type, net_id, seen_today in best_result.used_networks:
+            if seen_today:
                 # only add network if it was last_seen today
-                result_networks[network[0]].add(network[1])
+                result_networks[net_type].add(net_id)
```
willkg: Much clearer! :+1:

ichnaea/api/locate/cell.py
```diff
@@ -42,7 +43,7 @@
         ("age", numpy.int32),
         ("signalStrength", numpy.int32),
         ("score", numpy.double),
-        ("id", "S11"),
+        ("id_b64", "S16"),
```
willkg: I've never used numpy. This changes the field from a 0-terminated bytes array of 11 bytes to a 0-terminated bytes array of 16 bytes, right? If the mac address is in hex and we're not really doing math on it (it's just a label), why not switch it to a U? Is it just space concerns? Are there other niceties of being a 0-terminated byte array?

jwhitlock: The short answer is that cell IDs are more complex than mac addresses, and a binary field keeps the matching algorithm generic enough to handle all the station types.

Yes, `S16` is a bytestring of length 16. I came up with 16 by base64-encoding an 11-character string. The numpy type appears to be a null-terminated C-string, with all the issues that brings up. See the numpy issue for more weirdness.

The binary string is used to match the unique database key for the row. For WiFi and bluetooth, it is the binary encoding of the MAC ID. For cell stations and areas, it is the binary packed version of the type, the mobile country code (mcc), the mobile network code (mnc), the location area code (lac), and a cell identifier (cid):

https://github.com/mozilla/ichnaea/blob/14b73b16c702abb1a3f8262330631e56e521a938/ichnaea/models/cell.py#L222-L237

By using the base64 encoded binary ID, we avoid dropping training null characters, and still compare using database IDs in ``InternalPositionSource``.

jwhitlock:
Thanks!

PR #1012, opened 2019-12-13
Use a shared session for tests

jwhitlock:
When running tests, use ``scoped_session`` to ensure that test code, app code, and task code all work with the same session. This allows ``session.begin_nested()`` to be used when testing code that calls ``session.rollback()``. This is currently one test, the recently refactored test for deadlocks in ``StationUpdater``, but more are expected as we add more deadlock handlers.

There are a few changes to tests when using a shared, thread-local session. The first is that ``session.close()`` should only be called once at the end of tests, rather than once for each provisioned session. The function-scoped ``session`` fixture handles closing the session, and a new function ``db.release_session(session)`` can be used instead of ``session.close()``.

Because the ``session`` fixture is in charge of session teardown, it must be included if a new session is provisioned by tested code, or else the session will remain allocated and cause an error in the next test that uses a session. If ``db.session()`` is called without the ``session`` fixture, an exception is raised, so that the correct test fails.

This is follow-on work for issue #991

### DIFF
--- a/ichnaea/conftest.py
+++ b/ichnaea/conftest.py
@@ -153,7 +153,7 @@ def database():
 
 @pytest.fixture(scope="session")
 def db(database):
-    db = configure_db(uri=get_sqlalchemy_url())
+    db = configure_db(uri=get_sqlalchemy_url(), test=True)
     yield db
     db.close()
 
@@ -178,10 +178,11 @@ def restore_db(db):
 
 @pytest.fixture(scope="function")
 def session(db):
+    """Return a thread-local session, and clean up at test teardown."""
     with db.engine.connect() as conn:
         with conn.begin() as trans:
-            db.session_factory.configure(bind=conn)
-            session = db.session()
+            db.has_session_fixture = True
+            session = db.session(bind=conn)
 
             # Set the global session context for factory-boy.
             SESSION["default"] = session
@@ -190,8 +191,8 @@ def session(db):
 
             trans.rollback()
             session.close()
-            db.session_factory.configure(bind=db.engine)
-
+            db.session_factory.remove()
+            db.has_session_fixture = False
     API_CACHE.clear()
 
 
--- a/ichnaea/data/tests/test_station.py
+++ b/ichnaea/data/tests/test_station.py
@@ -97,8 +97,7 @@ class TestDatabaseErrors(BaseStationTest):
         )
         session.add(cell)
         session.commit()
-        # TODO: Find a more elegant way to do this
-        db.tests_task_use_savepoint = True
+        session.begin_nested()  # Protect test cell from task rollback
 
         error = errclass(errno, errmsg)
         wrapped = InterfaceError.instance(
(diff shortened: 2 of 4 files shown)


jwhitlock:
This addresses the TODO items on PR #1009. I tried a few approaches to get to this solution.

First, I read about SQLAlchemy sessions. Important sections were:
* [Session Basics](https://docs.sqlalchemy.org/en/13/orm/session_basics.html). It has been about 10 years since I've worked with SQLAlchemy, and I needed to relearn the concepts.
* [Transactions and Connection Management](https://docs.sqlalchemy.org/en/13/orm/session_transaction.html). This directly addresses the issues I had with rollbacks and tests, but the examples assume a simpler test environment than we use, and are based on ``unittest.TestCase``, so it doesn't directly map to our tests.
* [Contextual/Thread-local Sessions](https://docs.sqlalchemy.org/en/13/orm/contextual.html) goes into ``scoped_session`` and how to use it. I also learned about [zope.sqlalchemy](https://github.com/zopefoundation/zope.sqlalchemy) and [pyramid_tm](https://docs.pylonsproject.org/projects/pyramid_tm/en/latest/), frameworks that provide transaction support on top of SQLAlchemy. I'd like to research these frameworks rather than continuing to roll our own.
 
First, I tried to write various alternate fixtures, for example a ``shared_session`` fixture so that all tests would use the same session. There are some neat techniques used in [pytest-django](https://github.com/pytest-dev/pytest-django) that dynamically load alternate database implementations. For example, [the db fixture](https://github.com/pytest-dev/pytest-django/blob/4b83be9c5ad11415db805e1cdde06a75bf7ea013/pytest_django/fixtures.py#L193-L216) has code that says "if ``transactional_db`` is in the fixture list, use that instead". I tried going down this path, but I got stuck with the interaction between the session-scoped ``db`` fixture and the function-scoped ``session`` fixture.

The blocker to the alternate fixture approach was that lots of code calls ``db.session()``, bypassing the fixture. This code is mostly in the web app and task app code, which calls ``db_session`` and ``db_worker_session`` to provision an auto-closing session. There is some code in the ``session`` fixture to bind the ``db.session_factory`` to a transaction, so that new sessions are bound to the transaction and all the test changes are rolled back during teardown. However, these sessions are independent, and a ``begin_nested`` in one did not create a nested transaction in the others. ``db.session()`` would have to instead return the same session with each call, which is exactly what ``scoped_session`` is intended to do.

I tried a few variations that setup ``scoped_session`` inside a fixture, or inside ``Database.session``, and hit dead-ends, database timeouts, and other weird behaviour. Everything clicked when instead I moved ``scoped_session`` into ``Database.__init__`` instead, and required a much smaller set of changes. I then refactored a few times before arriving at ``TestDatabase``.

A further change would be to rename ``TestDatabase`` as ``ScopedDatabase`` or similar, and use the pytest fixture trick to use it only when a shared session is needed, such as adding ``db_shared_session`` to the fixture list. Only one test needs the shared session, and the other 816 work fine with shared or independent sessions. The application continues to uses indepedent sessions at runtime.  If this sounds useful, I'd like to explore it while all the code is still fresh in my mind.

jwhitlock:
I investigated the fixture-based idea for a bit. One issue is that, in order for the `db` fixture to be evaluated at runtime and swapped for one using ``scoped_session``, it needs to go from a session-scoped fixture to a function-scoped fixture. This is OK, but a bunch of session-scoped fixtures that build on `db` like `global_celery`, also go to function scope, and basically everything becomes function scoped.

I'd like to avoid it unless we hit a test that requires independent sessions.

jwhitlock:
I was able to get fixture-based shared sessions working in #1013, and I think that PR is a better solution than switching to shared sessions all the time.

PR #1013, opened 2019-12-17
Add shared session fixtures

jwhitlock:
*Note*: This is an alternate implementation of PR #1012. Notes about the differences are at the bottom. This is follow-on work for issue #991.
*Note 2*: Reworked and force pushed, new PR summary.

Previously, all tests used independent sessions bound to a transaction. This fails when a test needs to setup data in the test, and then tests code that issues a transaction rollback. This results in the setup data being rolled back as well. This is currently one test, the recently refactored test for deadlocks in ``StationUpdater``, but more are expected as we add more deadlock handlers.

The Database class now takes a ``shared=True`` keyword. When set, SQLAlchemy's ``scoped_session`` is used to return a thread-local session when ``db.session()`` is called. For these shared sessions, ``session.close()`` should be called once, for example in ``db.close()``, instead of by each caller of ``db.session()``. A new method ``db.release_session(session)`` helps callers release sessions the right way.

Move the existing database fixtures to new ones that make it explicit what kind of sessions are returned:

* ``db`` → ``db_independent_session``
* ``session`` → ``independent_session``
* ``global_app`` → ``app_independent_session``

Add some new fixtures that use a thread-local session, so that the test code, app, and celery can work in the same session:

* ``db_shared_session``
* ``shared_session``
* ``app_shared_session``
* ``celery_shared_session``

These function-scoped fixtures now return the shared session variants if ``db_shared_session`` or ``shared_session`` is requested, and the traditional independent sessions by default:

* ``db``
* ``session``
* ``app``
* ``celery``

Re-using these names means that most tests do not need changing to get the default independent session implementation. A few require additional fixtures to ensure session-scoped fixtures are setup.

This is the fixture alternative mentioned in https://github.com/mozilla/ichnaea/pull/1012#issuecomment-565611638. It was close enough that I wanted to finish it off, and it has grown on me since last week.

I then continued the work to use session-scoped fixtures when possible, so that there isn't (as much of) a speed penalty to using the just-in-time fixtures.

### DIFF
--- a/ichnaea/conftest.py
+++ b/ichnaea/conftest.py
@@ -17,7 +17,7 @@ from ichnaea.api.locate.searcher import (
     configure_region_searcher,
 )
 from ichnaea.cache import configure_redis
-from ichnaea.db import configure_db, create_db, get_sqlalchemy_url
+from ichnaea.db import Database, configure_db, create_db, get_sqlalchemy_url
 from ichnaea.geocode import GEOCODER
 from ichnaea.geoip import CITY_RADII, configure_geoip
 from ichnaea.http import configure_http_session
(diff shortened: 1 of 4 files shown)


willkg:
My cursory glance at your explanations gives me a gut feeling that this is the way to go. Sure, session scoping saves some time, but I prefer simpler interactions of things, I think the tests you're adding are important to add, and I'd rather go for the simpler and more expensive solution. I don't think the additional expense here is a big deal.

I think the third option is to have a separate test suite with the deadlock tests. We have two test suites in Socorro and run them back to back in `make test`. We could do something like that here, too.

jwhitlock:
I think the two test suites in Socorro are for the processor and the web app. That seems natural to me, they could be split into two different repos. This is for testing a database error case that is difficult to simulate, and I'm not ready to split up tests for that yet. So, I'm ruling out option 3 for now.

I'm going to continue with this fixture solution. I want to experiment for a few hours with some fixture scoping tricks to see if I can get down the time cost. I'll request a review when I'm done or give up.

jwhitlock:
I re-worked the fixtures to retain session scoping, and used fixture-scoped fixtures to pick between them as needed. This gets tests back to within a second of tests on master (24 seconds for me, 52s for the step in CircleCI).

I also integrated the ``shared=True`` code into the base ``Database`` class, which makes it easier to understand the differences between the sessions generated with `scoped_session` and the "traditional" sessions. There is some safety code for tests to make sure ``scoped_session`` is loaded when ``db_scoped_session`` is used, but this has moved to a derived class ``TestDatabase`` that lives with the test fixtures.

@willkg, I think this is ready to review.

jwhitlock:
Thanks!

PR #1014, opened 2019-12-17
StationUpdater: Use backoff.on_exception decorator

jwhitlock:
Add new dependency [backoff](https://github.com/litl/backoff), which provides a ``backoff.on_exception`` decorator that can be used to handle retrying for exceptions. This may help standardize the retry logic before we roll it out to other updater tasks.

I wasn't able to use ``backoff.on_exception`` as a traditional decorator, because the ``METRICS`` call needs access to both ``self.station_type`` and the error number from the exception. If I dropped the metrics call, or made it a counter without tags, then it could be simplified down to a decorator.

This is follow-on work for issue #991.

### DIFF
--- a/ichnaea/data/station.py
+++ b/ichnaea/data/station.py
@@ -1,7 +1,7 @@
 from collections import defaultdict
 from datetime import timedelta
-import time
 
+import backoff
 import markus
 import numpy
 from pymysql.err import MySQLError
@@ -400,6 +400,15 @@ class CellState(StationState):
         return values
 
 
+def is_retryable(exception):
+    """Return True if the exception is retryable."""
+    return (
+        isinstance(exception, StatementError)
+        and isinstance(exception.orig, MySQLError)
+        and exception.orig.args[0] in (LOCK_DEADLOCK, LOCK_WAIT_TIMEOUT)
+    )
+
+
 class StationUpdater(object):
 
     obs_model = None
@@ -407,9 +416,6 @@ class StationUpdater(object):
     station_type = None
     stat_obs_key = None
     stat_station_key = None
-    _retriable = (LOCK_DEADLOCK, LOCK_WAIT_TIMEOUT)
-    _retries = 3
-    _retry_wait = 1.0
 
     def __init__(self, task, shard_id=None):
         self.task = task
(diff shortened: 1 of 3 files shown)


willkg:
Do we look at that metric at all? I don't remember seeing it in the dashboards.

jwhitlock:
> Do we look at that metric at all? I don't remember seeing it in the dashboards.

It's a new stat, added in PR #1009. Previously (1.5.x branch), these errors were handled without being counted, but the handler broke with a SQLAlchemy update that changed the exception class. With the broken handler, it seemed there were 500 - 1000 uncaught deadlock errors per day.  I'm now measuring to 4000 - 6000 handled deadlocks (code 1213) per day, and another 50 or so lock timeouts (code 1205) a day. At first, it looked like there were less per day due to them being handled.

This change is a bit uglier than I'd like, and seems to make it more challenging to understand the code. Maybe I should try again, to make a decorator that looks good when used even if the implementation is a little uglier.

ichnaea/data/station.py
```diff
@@ -553,51 +559,50 @@ def __call__(self):
...
-                    self.queue_area_updates(pipe, updated_areas)
+        with self.task.db_session() as session:
+            for shard, shard_values in sharded_observations.items():
+                areas = self.update_shard(session, shard, shard_values, stats)
+                updated_areas.update(areas)
+        return updated_areas, stats
 
-                self.emit_stats(pipe, stats_counter)
+    def abort_or_count_exception(self, exception):
+        """Check if we should retry or abort on exception.
```
willkg: I'd rename this to `should_giveup` to continue the language that backoff is using. It's returning a boolean, but the method name suggests it's returning ABORT or COUNT and for me, it requires too much reading to figure out what's going on. I think generating metrics is really a side-effect of the method, so I don't think that needs to be part of the method name.

ichnaea/data/station.py
```diff
@@ -400,16 +400,22 @@ def submit_key(self):
         return values
 
 
+def is_retryable(exception):
+    """Return True if the exception is retryable."""
```
willkg: This should talk about what circumstance is covered. This isn't "is any exception retryable". This is more of a "is a SQLAlchemy exception retryable".

Maybe change the function name to `is_sqlalchemy_exception_retryable`? That's a bit of a mouthful. It's only used in one place, so maybe fold it into `should_giveup`?

jwhitlock:
As I'm writing this, I'm going back and forth between two approaches:

1. A targeted function that is specific to the needs of ``StationUpdater``
2. A generic method that could be extended to the needs of other updaters

I'm a big believer in "you ain't gonna need it", but I think #2 is where we are going. Should this PR change to add a retry loop to more updaters, or stay focused on this code to get ``backoff`` into the image?

jwhitlock:
I've moved the code to `_retry_updates_on_mysql_lock_fail`, which is specific to ``StationUpdater.update_observations``, but seems cleaner to me when implemented outside of the class. It no longer directly follows the decorator pattern, but that wasn't working well anyway, due to needing class and exception data at run time.

I can imagine a path from this to a retry wrapper on more updater methods:

* Move the function to ``ichnaea.data.retry`` or similar.
* Add more keyword arguments, such as ``metric_name=None`` and ``metric_tags=None``, to override behaviour for specific uses.

I've also moved the ``time.sleep`` mock to a fixture. I'm not convinced this is an improvement, but maybe it is closer to what you were hoping for.

ichnaea/data/tests/test_station.py
```diff
@@ -35,6 +35,13 @@
 from ichnaea import util
 
 
+@pytest.fixture
+def backoff_sleep_mock():
+    """Mock backoff's time.sleep, so that test remain fast."""
+    with mock.patch("backoff._sync.time.sleep") as mocked_sleep:
+        yield mocked_sleep
```
willkg: This is perfect!

jwhitlock:
Thanks!

PR #1015, opened 2019-12-17
StationUpdater: Use SELECT ... WITH UPDATE

jwhitlock:
When ``SELECT``ing station rows for updating, use ``WITH UPDATE`` to acquire a row lock. This may result in more lock timeouts, but less ``StaleDataError``s in stage and production. If this helps reduce issue #990, then we can roll out a similar change to the cell area and datamap update tasks.

### DIFF
--- a/ichnaea/data/station.py
+++ b/ichnaea/data/station.py
@@ -629,7 +629,9 @@ class StationUpdater(object):
 
 class MacUpdater(StationUpdater):
     def query_shard(self, session, shard, keys):
-        return (session.query(shard).filter(shard.mac.in_(keys))).all()
+        return (
+            (session.query(shard).filter(shard.mac.in_(keys))).with_for_update().all()
+        )
 
 
 class BlueUpdater(MacUpdater):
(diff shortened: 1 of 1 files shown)


jwhitlock:
I'd like to deploy this to stage, and see what happens. I'm expecting:

* ``StaleDataError`` count to go to zero for the ``StationUpdate`` tasks (wifi, cell, and bluetooth). I suspect that is caused by multiple tasks updating the same station at the same time, and this would cause them to wait until each other are done.
* Deadlock errors to drop or go to zero for the ``StationUpdate`` tasks. I believe the deadlocks are caused by two tasks having an overlap in the set of updated stations, but instead of deadlocks they will wait for the lock, which means...
* More lock timeout errors for ``StationUpdate`` tasks. Now the tasks will wait for each other if they are looking at the same rows, but may timeout while waiting for the lock.

I could do this before the winter break, and we'd get a couple of weeks of data. Only a few hours is needed to see the effect however, since the ``StaleDataError`` happens many times an hour. I'm OK waiting until after break as well, so that master is closer to a "ready to deploy" state, in case of holiday emergencies.

@willkg, ready for review, and for your opinion on merge now or 2020.

jwhitlock:
Thanks!

PR #1016, opened 2019-12-19
Add `make update-vendored`, and update vendor libraries and geo databases

jwhitlock:
There are two vendor libraries built from source included in the repo:

* ``vendor/datamaps-76e620adabbedabd6866b23b30c145b53bae751e.tar.gz`` - source for data maps tools to generate tiles of contributions, from https://github.com/ericfischer/datamaps
* ``vendor/libmaxminddb-1.3.2.tar.gz`` - source for the MaxMind database library, needed for the C extension variant of [maxmindb](https://pypi.org/project/maxminddb/), from https://github.com/maxmind/libmaxminddb

There are three data files used in development from https://github.com/maxmind/MaxMind-DB/:
* ``ichnaea/tests/data/GeoIP2-City-Test.json`` - source data for the city database
* ``ichnaea/tests/data/GeoIP2-City-Test.mmdb`` - Binary city database, for IP-based lookups
* ``ichnaea/tests/data/GeoIP2-Connection-Type-Test.mmdb`` - Binary connection-type database, for testing system response to an invalid GeoIP database

The age of the geo database is shown at the ``__heartbeat__`` endpoint, and there is a test that this database is less than 1000 days old.

This PR adds a `make update-vendored` command to refresh all these files at once. The data maps project has not been updated, but the MaxMind library is now at 1.4.2. The test databases are also updated, and go from 232 to 45 days old when I visit ``http://localhost:8000/__heartbeat__``.


Makefile
```diff
@@ -44,6 +44,7 @@ default:
 	@echo "  docs             - generate Sphinx HTML documentation, including API docs"
 	@echo "  buildjs          - generate js static assets"
 	@echo "  buildcss         - generate css static assets"
+	@echo "  download         - re-download vendor source and test data"
```
willkg: While this is downloading things, I think the point of it is more helpful for a make target. It's updating vendored things that age over time. I wonder if there's a better name for this make target. Maybe "update-vendored" or something like that?

Makefile
```diff
@@ -131,3 +132,7 @@ runservices: my.env .docker-build
 .PHONY: stop
 stop: my.env
 	${DC} stop
+
+.PHONY: download
+download: my.env
+	${DC} run --rm --no-deps app shell make -f docker.make download
```
willkg: As a side note, I've never liked the other make files. I think they're currently less makey and more hard-to-debug shell scripts. At some point, I want to look at redoing them.

This is fine for now, but if you feel like it was ornery (you can't run shellcheck on make files), then maybe we should split them into shell scripts in the `docker/` or `bin/` directories.

docs/local_dev.rst
```diff
@@ -250,3 +250,20 @@ it is available in that local directory.
 
 To view the documentation open ``file://docs/build/html/index.html``
 with a web brower.
+
+Updating Test GeoIP Data and Libraries
+======================================
+The development environment uses a test MaxMind GeoIP database, and a test
+confirms that this is less than 1000 days old. To update this database and
```
willkg: I know because I read the code that the test is part of Ichnaea, but I think this isn't clear in the text. I think I'd tweak it to "and an Ichnaea test" or "a test in the Ichnaea test suite" or something like that.

ichnaea/tests/data/README.rst
```diff
@@ -8,3 +8,5 @@ To view a copy of this license, visit
 http://creativecommons.org/licenses/by-sa/3.0/ or send a letter to
 Creative Commons, 444 Castro Street, Suite 900, Mountain View,
 California, 94041, USA
+
+To download fresh copies and confirm they work, run ``make download test``.
```
willkg: This is super helpful! Thank you!

ichnaea/webapp/tests.py
```diff
@@ -62,6 +62,7 @@ def test_ok(self, app):
             assert isinstance(data[name]["time"], int)
             assert data[name]["time"] >= 0
 
+        # Run "make download" when this gets too old
```
willkg: This is super helpful, too! Thank you!

Makefile
jwhitlock: let me see if I can convert ``docker.make`` quickly...

jwhitlock: The issues I saw in the first 30 minutes:

* It is easy to set `DATAMAPS_COMMIT` and `LIBMAXMIND_VERSION` in one place in `docker.make`, and then use it in multiple places. It's not clear where these should be set when refactoring to scripts.
* When building the Docker image, the Dockerfile copies ``docker.make`` to the image and runs it. When using scripts, should the scripts be copied one at a time, or should we copy the whole `docker` folder?

I'm guessing you have strong opinions on what these scripts should look like, and how they should be called. I'd rather review your refactor than try to implement it myself.

jwhitlock:
Added a commit to:

* Changed from `make download` to `make update-vendored`
* Clarified that the Ichnaea test suite fails if the test geo database is too old

I tried to migrate from `docker.make` to shell scripts, but I think this deserves its own PR, and I think @willkg should do it 😀 Thanks for the review!

PR #1022, opened 2020-01-06
bug 1607327: switch to dependabot

willkg:
With pyup, we were looking at Python dependencies monthly. This keeps that and adds Docker images, too.

This file is "valid" configuration per the dependabot configuration validator:

https://dependabot.com/docs/config-file/validator/

First step is to add the config file. Then I can add it to dependabot via their website. If there are any issues, we can fix them as we move forward.


willkg:
@jwhitlock We can probably add JS files, too. Do we want to do that?

jwhitlock:
> @jwhitlock We can probably add JS files, too. Do we want to do that?

Yeah probably, although I'm not looking forward to reviewing JS updates for another project.

willkg:
That's a super point--I'll skip adding JS for now.

PR #1056, opened 2020-01-08
bug 1607906: Remove string handling from encode/decode_gzip

jwhitlock:
For ``util.encode_gzip``, drop optional ``encoding`` parameter and require that the input is bytes. This simplifies half of the callers, and calling ``.encode()`` on the string data does the right thing for the rest.

For ``utils.decode_gzip``, drop the optional ``encoding`` parameter and return the decompressed bytes instead of a string decoded from UTF-8. This simplifies most callers, who were sending ``encoding=None`` to get this behavior.

Also, use Python 3.2's ``gzip.compress`` and ``gzip.decompress``, which take care of the details of feeding bytes through ``GzipFile``. However, errors in the bytes are still raised as if they come from a file, like ``gzip.BadGzipFile``, so the exception handler remains complex.

### DIFF
--- a/ichnaea/util.py
+++ b/ichnaea/util.py
@@ -1,7 +1,6 @@
 from contextlib import contextmanager
 from datetime import datetime
 import gzip
-from io import BytesIO
 from itertools import zip_longest
 import json
 import os
@@ -34,12 +33,7 @@ def gzip_open(filename, mode, compresslevel=6):
 
 def encode_gzip(data, compresslevel=6):
     """Encode the passed in data with gzip."""
-    out = BytesIO()
-    with gzip.GzipFile(
-        None, "wb", compresslevel=compresslevel, fileobj=out
-    ) as gzip_file:
-        gzip_file.write(data)
-    return out.getvalue()
+    return gzip.compress(data, compresslevel=compresslevel)
 
 
 def decode_gzip(data):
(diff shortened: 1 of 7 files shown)


ichnaea/util.py
```diff
@@ -32,29 +31,18 @@ def gzip_open(filename, mode, compresslevel=6):
...
     """Encode the passed in data with gzip."""
-    if encoding and isinstance(data, str):
-        data = data.encode(encoding)
-    out = BytesIO()
-    with gzip.GzipFile(
-        None, "wb", compresslevel=compresslevel, fileobj=out
-    ) as gzip_file:
-        gzip_file.write(data)
-    return out.getvalue()
+    return gzip.compress(data, compresslevel=compresslevel)
```
jwhitlock: This could now be:

```
encode_gzip == gzip.compress
```

which would add access to the ``mtime`` parameter. The gzip format includes a 4-byte modification time (see [wikipedia](https://en.wikipedia.org/wiki/Gzip)), and [the ``mtime`` parameter](https://docs.python.org/3/library/gzip.html#gzip.GzipFile.mtime) allows you to pick a time rather than the current instant, so that you can get the same exact bytes in two different runs. However, this only impacts a test, which already has a workaround. 

Also, [someone in 2011](https://stackoverflow.com/a/4928651/10612) didn't like the quality of the ``gzip`` module, and thought it wasn't as good as the other standard library stuff.

I now know too much about gzip.

jwhitlock:
Thanks!

PR #1057, opened 2020-01-09
Expand lock retries and SELECT ... WITH UPDATE to cell area and datamap updates 

jwhitlock:
Move ``_retry_updates_on_mysql_lock_fail``  to ``retry_on_mysql_lock_fail`` in ``ichnaea.db``, move station-specific stuff to parameters, and implement as a standard decorator.

This makes it a little more awkward to use in ``StationUpdater``, but allows it to be expanded to cell area and datamap updating. This addresses issue #991.

The ``data.station.dberror`` metric may have tag ``type=cellarea`` for retryable
exceptions in ``CellAreaUpdater``, and a new metric ``datamaps.dberror`` is
incremented from ``DataMapUpdater``.

Add a ``WITH UPDATE`` clause to the ``SELECT`` for cell areas and datamaps, to avoid ``StaleDataErrors``s, and maybe turn some into deadlocks (which will now be retried). This addresses issue #990.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -36,6 +36,7 @@ Metric Name                      Type    Tags
 `data.station.dberror`_          counter type, errno
 `data.station.new`_              counter type
 `datamaps`_                      timer   func, count
+`datamaps.dberror`_              counter errno
 `locate.fallback.cache`_         counter fallback_name, status
 `locate.fallback.lookup`_        counter fallback_name, status
 `locate.fallback.lookup.timing`_ timer   fallback_name, status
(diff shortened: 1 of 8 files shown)


jwhitlock:
The collisions that require these changes only happen in full deployment environments like stage and prod, so I don't think it makes sense to try to trigger the code in the development environment. I think tests passing is as far as we can go for local validation.

In stage, there should be no more ``StaleDataError``s reported to Sentry. The number of ``OperationalError``s from deadlocks should be reduced, since they will now be retried. There may be 10 or less ``OperationalError``s per day, representing persistent failures after 3 retries, down from hundreds per hour.

A lot happened at once, which may not be obvious when reading the changes from top to bottom:

* The pytest fixture ``backoff_sleep_mock`` moved from ``ichnaea/data/tests/test_station.py`` to ``ichnaea/conftest.py``, without changes.
* The helper ``_retry_updates_on_mysql_lock_fail`` in ``ichnaea/data/tests/test_station.py`` was recreated as a more traditional function decorator ``retry_on_mysql_lock_fail`` in ``ichnaea/db.py``
* ``retry_on_mysql_lock_fail`` is fully tested in `` ichnaea/tests/test_db.py``. The retriable exceptions are defined here as ``RETRIABLES``, and imported in ``ichnaea/data/tests/test_station.py`` for the existing test.
* The "inner update" of ``CellAreaUpdater`` and ``DataMapUpdater`` are refactored from ``__call__`` into their own class methods, so that the decorator ``retry_on_mysql_lock_fail`` can be applied.

jwhitlock:
Thanks!

PR #1058, opened 2020-01-14
991: Lock cellarea rows early to avoid deadlocks

jwhitlock:
With PR #1057 deployed to stage, there are more deadlocks than I expected when updating the cellarea tables. My guess is that this is because the ``SELECT ... FOR UPDATE`` is done inside the loop, so that two tasks working on the same rows may get partially through processing before they try to update the same rows and deadlock.

This PR locks all the cellarea rows at the start of processing, so that overlapping tasks will pause when getting a lock, making them sequential. It makes cellarea row locking more like station row locking.

An alternate implementation would process the IDs in sorted order. The lock contention would happen when the overlap occurs, but the risk of deadlock would be eliminated. However, in some cases this would result in some work occurring (selecting and processing a bunch of cell data) followed by a lock timeout and rollback. I'd prefer to let a task complete its work, then start the next task.

Part of issue #991.

### DIFF
--- a/ichnaea/data/area.py
+++ b/ichnaea/data/area.py
@@ -32,9 +32,20 @@ class CellAreaUpdater(object):
     )
     def update_areas(self, areaids):
         """Update the cell areas based on cell station records."""
+        areaid_set = set(areaids)
         with self.task.db_session() as session:
-            for areaid in set(areaids):
-                self.update_area(session, areaid)
+            # Fetch and lock existing cellarea rows
+            rows = session.execute(
+                select([self.area_table.c.areaid])
+                .where(self.area_table.c.areaid.in_(areaid_set))
+                .with_for_update()
+            ).fetchall()
+            existing_areas = set(row[0] for row in rows)
+
+            # Update each cellarea row from cell tables
+            for areaid in areaid_set:
+                area_is_new = decode_cellarea(areaid) not in existing_areas
+                self.update_area(session, areaid, area_is_new)
 
     def region(self, ctr_lat, ctr_lon, mcc, cells):
         region = None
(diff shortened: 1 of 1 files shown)


ichnaea/data/area.py
```diff
@@ -102,24 +113,6 @@ def update_area(self, session, areaid):
             return
 
         # Otherwise update the area entry based on all the cells
-        area = session.execute(
```
jwhitlock: This query fetched the existing cellarea data, but was only used to determine if it already existed (use ``UPDATE``) or should be created (use ``INSERT``).

ichnaea/data/area.py
```diff
@@ -164,7 +157,7 @@ def update_area(self, session, areaid):
         if cell_last_seen:
             last_seen = max(cell_last_seen)
 
-        if area is None:
+        if area_is_new:
```
jwhitlock: I preferred to use ``area_exists``, but ``area_is_new`` makes this existing statement more natural, and keeps it a one-line change.

ichnaea/data/area.py
```diff
@@ -32,9 +32,20 @@ def __call__(self):
...
+            rows = session.execute(
+                select([self.area_table.c.areaid])
+                .where(self.area_table.c.areaid.in_(areaid_set))
+                .with_for_update()
+            ).fetchall()
+            existing_areas = set(row[0] for row in rows)
+
+            # Update each cellarea row from cell tables
+            for areaid in areaid_set:
+                area_is_new = decode_cellarea(areaid) not in existing_areas
```
jwhitlock: ``areaid`` is the binary packed string from the Struct ``CELLAREA_STRUCT``. The SQLAlchemy ``select`` above decodes IDs to a tuple, so ``decode_cellarea`` is needed for comparison.

jwhitlock:
Thanks!

PR #1059, opened 2020-01-16
bug 1607906: re-write preprocess_request for detailed errors

jwhitlock:
PR #1028 unintentionally removed an exception handler for string decoding issues. Previously, bytes were decoded as part of ``json.loads``:

https://github.com/mozilla/ichnaea/blob/cc3d9f314110c4acb9af6161f9998b7f06490048/ichnaea/api/views.py#L106-L110

If decoding failed, a ``UnicodeDecodeError`` was raised, which is derived from ``ValueError``, so it was caught and handled.

In Python 3.8, ``json.loads`` no longer uses the `encoding` parameter, so bytes decoding was moved to a separate step, without a ``try``/``except`` block. Some callers have triggered this, raising a handful of exceptions in production (see [bug 1607906](https://bugzilla.mozilla.org/show_bug.cgi?id=1607906#c2)). The most common root cause appears to be sending a truncated gzip-compressed payload, which resulted in an unsuccessful decode of a compresses bytestring.

This PR makes a number of changes to ``preprocess_request`` to solve this issue:

* Add tests for truncated gzip-compressed content and badly encoded content
* Allow ``ParseError`` to take an optional ``details`` parameter at initialization, which is added to the response JSON. Expand the exception tests to test instances as well as exception classes.
* Add a ``detail`` section to the ``ParseError`` JSON, which gives the caller more information, so they can tell if it is a gzip, unicode, or JSON decoding error, or a payload validation error.
* Simplify the return value of ``preprocess_request``, since the ``error`` element of the tuple is unused.

### DIFF
--- a/ichnaea/api/exceptions.py
+++ b/ichnaea/api/exceptions.py
@@ -152,6 +152,17 @@ class ParseError(BaseAPIClientError):
     domain = "global"
     reason = "parseError"
     message = "Parse Error"
+    error_details = None
+
+    def __init__(self, details=None):
+        self.error_details = details
+        super().__init__()
+
+    def json_body(self):
+        content = super().json_body()
+        if self.error_details:
+            content["details"] = self.error_details
+        return content
 
 
 class ServiceUnavailable(BaseAPIServiceError):
--- a/ichnaea/api/exceptions.py
+++ b/ichnaea/api/exceptions.py
@@ -34,8 +34,7 @@ class JSONException(HTTPException):
     def __str__(self):
         return "<%s>: %s" % (self.__class__.__name__, self.code)
 
-    @classmethod
-    def json_body(cls):
+    def json_body(self):
         """A JSON representation of this response."""
         return {}
 
@@ -46,11 +45,6 @@ class UploadSuccess(JSONException):
     """
 
     code = 200
-
-    @classmethod
-    def json_body(cls):
-        """A JSON representation of this response."""
-        return {}
 
 
 class UploadSuccessV0(UploadSuccess):
@@ -74,16 +68,19 @@ class BaseAPIError(JSONException):
     reason = ""
     message = ""
 
-    @classmethod
-    def json_body(cls):
+    def json_body(self):
         """A JSON representation of this response."""
         return {
             "error": {
                 "errors": [
-                    {"domain": cls.domain, "reason": cls.reason, "message": cls.message}
+                    {
+                        "domain": self.domain,
+                        "reason": self.reason,
+                        "message": self.message,
+                    }
                 ],
-                "code": cls.code,
-                "message": cls.message,
+                "code": self.code,
+                "message": self.message,
             }
         }
 
--- a/ichnaea/api/locate/tests/base.py
+++ b/ichnaea/api/locate/tests/base.py
@@ -208,7 +208,9 @@ class BaseLocateTest(object):
     def check_queue(self, data_queues, num):
         assert data_queues["update_incoming"].size() == num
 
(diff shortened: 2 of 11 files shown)


ichnaea/api/submit/tests/test_submit_v2.py
```diff
@@ -265,6 +265,11 @@ def test_error(self, app, celery, raven):
             ],
             status=400,
         )
+        assert response.json["details"]["validation"] == {
+            "items.0.wifiAccessPoints.0.macAddress": (
+                "10 is not a string: {'macAddress': ''}"
```
jwhitlock: Here's what a "real" validation error looks like. It's kind of ugly.

ichnaea/api/exceptions.py
```diff
@@ -152,6 +152,17 @@ class ParseError(BaseAPIClientError):
...
+
+    def __init__(self, details=None):
+        self.error_details = details
+        super().__init__()
+
+    def json_body(self):
+        json = super().json_body()
+        if self.error_details:
+            json["details"] = self.error_details
+        return json
```
willkg: I probably wouldn't use the variable name `json` since if this file ever imports the `json` module, then that gets problematic. However, it looks like other parts of this file use `json` as a variable name, so meh.

ichnaea/api/exceptions.py
```diff
@@ -152,6 +152,17 @@ class ParseError(BaseAPIClientError):
     domain = "global"
     reason = "parseError"
     message = "Parse Error"
+    error_details = None
+
+    def __init__(self, details=None):
+        self.error_details = details
+        super().__init__()
+
+    def json_body(self):
```
willkg: This is shadowing a class method with an instance method. That's problematic. I think it's probably better to make `json_body` an instance method in all these classes and adjust the tests accordingly.

willkg: Probably worth fixing `ParseError` which is also shadowing a class method with an instance method.

ichnaea/api/locate/tests/base.py
```diff
@@ -445,6 +457,32 @@ def test_gzip(self, app, data_queues):
...
+            "GZIPDecodeError(\"EOFError('Compressed file ended before the"
+            " end-of-stream marker was reached')\")"
+        )
+        self.check_response(data_queues, res, "parse_error", details={"decode": detail})
+
+    def test_bad_encoding(self, app, data_queues):
+        body = b'{"comment": "R\xe9sum\xe9 from 1990", "items": []}'
+        assert "Résumé" in body.decode("iso8859-1")
+        with pytest.raises(UnicodeDecodeError):
+            body.decode("utf-8")
```
willkg: I like that you assert the test is set up properly. :+1:

ichnaea/api/exceptions.py
jwhitlock: I agree, it can be problematic. As long as I'm touching it, I should touch them all at once.

ichnaea/api/exceptions.py
jwhitlock: There's a lot of code that relies on ``json_body`` being a class method, but I think most of it is test code. I'll give this a shot.

jwhitlock:
Force-push to rebase on master, and:

* Change ``ParseError.json_body`` to use ``content`` instead of ``json`` as instance variable.  This new code is the only code that used ``json`` as a variable name.
* Add a commit to convert the rest of the ``JSONException`` exception tree to define ``json_body`` as an instance rather than a class method. This required changing some test code, but it is only used in ``JSONException.__init__`` and works as easily as an instance variable, so no application code needs to be updated.

PR #1073, opened 2020-02-05
Observation docs: center on reports, not signal

jwhitlock:
The Ichnaea station model attempts to determine the center point and the area range from which a station will be observed. This updates the document to reflect this, rather than suggest it is attempting to locate the station signal source. It includes a diagram of the observation update process, and the relationship between the model and the real-world station for some common use cases.

There are additional minor updates from a full read-through of the document.


jwhitlock:
My guide for these changes was comments by @hannosch on the first version: https://github.com/mozilla/ichnaea/pull/995#issuecomment-554797072. There's other interesting info about Ichnaea that didn't make sense for this document.

docs/algo/observations.rst
```diff
@@ -5,68 +5,171 @@ Observations
...
-of the observations, and further observations will result in a more accurate
-estimate of the station position.
+The Station Model
+=================
+
+A station, regardless of type, is modeled as a circle, where the center is the
+weighted average position of observations, and the radius is large enough to
+contain historical observations.  A model can be composed of observations from
+location queries, giving an estimated position and region for the station, but
+keeping it from being using in location queries. A model can instead be
```
willkg: I can't parse this sentence:

>  A model can be composed of observations from location queries, giving an estimated position and region for the station, but keeping it from being using in location queries.

Maybe the last part should be "from being *used* in location queries"? Even then, I'm not sure I understand the sentence.

jwhitlock: Yep, that doesn't make sense as written.

The concept is a little hard to convey, maybe I should just delete these sentences. The idea is that stations can be seen in location queries or in submission reports. 

In a location query, the caller is sending stations and asking for the position. If the query contains stations we know about and ones we don't, then we learned about a new station, and approximately where it is. We probably don't want to use this station's estimated position for future location queries, but I'm not sure yet what the code actually does. The station is tracked in the database, along with a region, but flagged as not coming from a GPS-backed position.

In a submission report, the caller provides a location from GPS or other GNSS source. We can then start building a model based on GPS-backed observations, and use these for future location queries to estimate a position.

There are stations without a GPS-backed position, and new non-GPS observations (from location queries) can adjust the estimated position. When a GPS-backed position appears, this is discarded, a new position is estimated, and only GPS-backed positions are used in the future. If a non-GPS observation comes in, it can no longer adjust the position, but can update the "last seen" date.

It's probably too complex for the first paragraph, and I should update it once I know for sure how queries work. Maybe by then I'll have a good short name to distinguish between the two types of observations and stations.

jwhitlock: It made a little more sense in the "Sources and Batching" section, and kept the intro tight. New commit added.

willkg:
Nice job!

Issue #468, opened 2015-11-11
Enable CORS on /v1/country endpoint


jbuck:
Hi there! We'd like to use Ichnaea for auto-detecting country on donate.mozilla.org. The only blocker to doing this would be to enable CORS on the endpoint so that visitors to donate.m.o could successfully complete the request.

Easy way to test in your browser whether it's working successfully or not:

`fetch("https://location.services.mozilla.com/v1/country").then(function(response) { console.log(response); }).catch(function(ex) { console.log(ex); });`

hannosch:
Ah, right. I'll get to this shortly. So far we have only been used from inside the guts of Firefox Desktop, not directly from web pages.

jbuck:
@hannosch Great, thank you!

hannosch:
Got most of the functionality in place, hoping to get this down and to stage this week, if all things go well it'll land in early production next week.

jbuck:
@hannosch awesome! If you'd like us to test staging for you, just mention me when that's deployed

hannosch:
@jbuck can you try and see if https://location.stage.mozaws.net/v1/country works for you now?

jbuck:
@hannosch Working great with CORS!

`fetch("https://location.stage.mozaws.net/v1/country").then((response) => { return response.json(); }).then((json) => { console.log(json); }).catch((ex) => { console.log(ex); }); `

PR #1075, opened 2020-02-10
bug 1610015: Restrict CORS to API endpoints

jwhitlock:
Restrict Cross-Origin Resource Sharing, via ``Access-Control-Allow-Origin`` header, to the API endpoints. Previously, CORS was allowed on all views. 

CORS for ``/v1/country`` was requested in November 2015 in https://github.com/mozilla/ichnaea/issues/468. This is appropriate for an API called from in-browser JavaScript from unknown sites, but it doesn't need to be applied to the homepage or other content pages.

### DIFF
--- a/ichnaea/webapp/tests.py
+++ b/ichnaea/webapp/tests.py
@@ -99,8 +99,6 @@ class TestHeartbeatErrors(object):
             "time": 0,
             "alembic_version": "unknown",
         }
-        assert res.headers["Access-Control-Allow-Origin"] == "*"
-        assert res.headers["Access-Control-Max-Age"] == "2592000"
 
     def test_geoip(self, broken_app):
         res = broken_app.get("/__heartbeat__", status=503)
@@ -113,9 +111,16 @@ class TestHeartbeatErrors(object):
         assert res.json["redis"] == {"up": False, "time": 0}
 
     def test_lbheartbeat(self, broken_app):
+        """LBHeartbeat returns OK for a broken application."""
         res = broken_app.get("/__lbheartbeat__", status=200)
         assert res.content_type == "application/json"
         assert res.json["status"] == "OK"
+
+    def test_no_cors_headers(self, broken_app):
+        """Heartbeat does not allow CORS."""
+        res = broken_app.get("/__heartbeat__", status=503)
+        assert "Access-Control-Allow-Origin" not in res.headers
+        assert "Access-Control-Max-Age" not in res.headers
 
 
 class TestLBHeartbeat(object):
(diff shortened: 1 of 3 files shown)


ichnaea/webapp/tests.py
```diff
@@ -113,18 +111,23 @@ def test_redis(self, broken_app):
         assert res.json["redis"] == {"up": False, "time": 0}
 
     def test_lbheartbeat(self, broken_app):
+        """LBHeartbeat returns OK for a broken application."""
```
jwhitlock: I added this comment because my first instinct was to delete this test ("Why is ``__lbheartbeat__`` tested in the ``__heartbeat__`` tests?). The purpose of the test is to show that ``__lbheartbeat__`` returns "healthy" even for a mis-configured application. If it did not, AWS may stop sending traffic, and a auto-scaling group may reboot the web head, making it more difficult to diagnose the root issue.

ichnaea/webapp/tests.py
```diff
@@ -99,8 +99,6 @@ def test_database(self, broken_app):
             "time": 0,
             "alembic_version": "unknown",
         }
-        assert res.headers["Access-Control-Allow-Origin"] == "*"
-        assert res.headers["Access-Control-Max-Age"] == "2592000"
```
jwhitlock: I moved these to their own test, to split it from the ``GET`` response.

ichnaea/webapp/tests.py
```diff
@@ -136,24 +139,16 @@ def test_post(self, app):
         assert res.content_type == "application/json"
         assert res.json["status"] == "OK"
 
-    def test_options(self, app):
```
jwhitlock: The service only supports ``OPTIONS`` for endpoints that need to support [pre-flight CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Preflighted_requests), so this test now fails.

jwhitlock:
You can test this in the Firefox debug console on a site without a CSP policy (like https://developer.mozilla.org), borrowing from https://github.com/mozilla/ichnaea/issues/468#issuecomment-157807528 .

This returns JSON now, but should be an error:

```
fetch("https://location.stage.mozaws.net/__heartbeat__").then((response) => { return response.json(); }).then((json) => { console.log(json); }).catch((ex) => { console.log(ex); });
```

This shows the expected error (``Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost:8000/__heartbeat__. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).``) when running against this PR:

```
fetch("http://localhost:8000/__heartbeat__").then((response) => { return response.json(); }).then((json) => { console.log(json); }).catch((ex) => { console.log(ex); });
```

When ``KEY`` is valid, this returns a 404 not found (no region for a local IP), but not a CORS error:

```
fetch("http://localhost:8000/v1/country?key=KEY").then((response) => { return response.json(); }).then((json) => { console.log(json); }).catch((ex) => { console.log(ex); });
```

When ``KEY`` is valid, this returns a region, and should after the PR lands as well

```
fetch("https://location.stage.mozaws.net/v1/country?key=KEY").then((response) => { return response.json(); }).then((json) => { console.log(json); }).catch((ex) => { console.log(ex); });
```

jwhitlock:
Thanks!

PR #1076, opened 2020-02-11
bug 1613233: Drop reports with null positions

jwhitlock:
In the ``InternalExporter``, ensure that a position of null/``None`` is dropped without an error. For other exporters, confirm that these reports are forwarded without errors.

In [v2/geosubmit](https://ichnaea.readthedocs.io/en/latest/api/geosubmit2.html), the position item is a dictionary that includes the latitude and longitude, position source, and other details. It is not
required by the geosubmit endpoints, and it is possible for clients to submit WiFi and other data without including a GPS-derived position.

This fixes [bug 1613233](https://bugzilla.mozilla.org/show_bug.cgi?id=1613233), where a null position was causing an exception to be raised.

### DIFF
--- a/ichnaea/data/export.py
+++ b/ichnaea/data/export.py
@@ -329,7 +329,8 @@ class InternalTransform(object):
         cells = self._parse_list(item, report, self.cell_id, self.cell_map)
         wifis = self._parse_list(item, report, self.wifi_id, self.wifi_map)
 
-        gps_age = item.get("position", {}).get("age", 0)
+        position = item.get("position") or {}
+        gps_age = position.get("age", 0)
         timestamp = item.get("timestamp")
         if timestamp:
             # turn timestamp into GPS timestamp
--- a/ichnaea/data/tests/test_export.py
+++ b/ichnaea/data/tests/test_export.py
@@ -36,6 +36,7 @@ class BaseExportTest(object):
         lat=None,
         lon=None,
         source=None,
+        set_position=True,
     ):
         reports = []
         timestamp = int(time.time() * 1000)
@@ -43,16 +44,19 @@ class BaseExportTest(object):
             pos = CellShardFactory.build()
             report = {
                 "timestamp": timestamp,
-                "position": {},
+                "position": None,
                 "bluetoothBeacons": [],
                 "cellTowers": [],
                 "wifiAccessPoints": [],
             }
-            report["position"]["latitude"] = lat or pos.lat
-            report["position"]["longitude"] = lon or pos.lon
-            report["position"]["accuracy"] = 17.0 + i
-            if source is not None:
-                report["position"]["source"] = source
+            if set_position:
+                report["position"] = {
+                    "latitude": lat or pos.lat,
+                    "longitude": lon or pos.lon,
+                    "accuracy": 17.0 + i,
+                }
+                if source is not None:
+                    report["position"]["source"] = source
 
             blues = BlueShardFactory.build_batch(blue_factor, lat=pos.lat, lon=pos.lon)
             for blue in blues:
@@ -89,7 +93,7 @@ class BaseExportTest(object):
         items = [
             {
                 "api_key": api_key,
-                "source": rep["position"].get("source", "gnss"),
+                "source": (rep["position"] or {}).get("source", "gnss"),
                 "report": rep,
             }
             for rep in reports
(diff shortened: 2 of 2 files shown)


ichnaea/data/export.py
```diff
@@ -329,7 +329,8 @@ def __call__(self, item):
         cells = self._parse_list(item, report, self.cell_id, self.cell_map)
         wifis = self._parse_list(item, report, self.wifi_id, self.wifi_map)
 
-        gps_age = item.get("position", {}).get("age", 0)
+        position = item.get("position") or {}
+        gps_age = position.get("age", 0)
```
jwhitlock: This is the actual fix for bug 1613233.

ichnaea/data/tests/test_export.py
```diff
@@ -36,23 +36,27 @@ def add_reports(
...
                 "bluetoothBeacons": [],
                 "cellTowers": [],
                 "wifiAccessPoints": [],
             }
-            report["position"]["latitude"] = lat or pos.lat
-            report["position"]["longitude"] = lon or pos.lon
-            report["position"]["accuracy"] = 17.0 + i
-            if source is not None:
-                report["position"]["source"] = source
+            if set_position:
```
jwhitlock: This is the code that creates reports for the export tests. The new parameter ``set_position`` defaults to ``True``, and provides the existing functionality. If set to ``False``, the report has ``"position": None`` and triggers the bug.

ichnaea/data/tests/test_export.py
```diff
@@ -736,3 +766,15 @@ def test_datamap(self, celery, session):
         self._update_all(session, datamap_only=True)
         assert celery.data_queues["update_datamap_ne"].size() == 1
         assert celery.data_queues["update_datamap_sw"].size() == 1
+
+    def test_no_position(self, celery, session, metricsmock):
```
jwhitlock: This is the test for the new ``position: None`` functionality. Previously, this would raise an exception.

ichnaea/data/tests/test_export.py
```diff
@@ -154,10 +168,10 @@ def mock_send(self, data, num=num):
 
 class TestGeosubmit(BaseExportTest):
     def test_upload(self, celery, session, metricsmock):
-        ApiKeyFactory(valid_key="e5444-794")
+        ApiKeyFactory(valid_key="e5444-7946")
```
jwhitlock: I changed this to match the ``api_key`` in the report. Its not clear how this worked before.

ichnaea/data/tests/test_export.py
```diff
@@ -167,6 +181,7 @@ def test_upload(self, celery, session, metricsmock):
         reports.extend(self.add_reports(celery, 1, source="gnss"))
         reports.extend(self.add_reports(celery, 1, api_key="e5444e9f-7946"))
         reports.extend(self.add_reports(celery, 1, api_key=None, source="fused"))
+        reports.extend(self.add_reports(celery, 1, set_position=False))
```
jwhitlock: This test confirms the `geosubmit` exporter does not have an issue with `position: None`.

ichnaea/data/tests/test_export.py
```diff
@@ -213,12 +228,14 @@ def test_upload(self, celery, session, metricsmock):
...
         )
         ApiKeyFactory(valid_key="e5444-794")
+        ApiKeyFactory(valid_key="no-position")
         session.flush()
 
         reports = self.add_reports(celery, 3)
         self.add_reports(celery, 3, api_key="e5444-794", source="gnss")
         self.add_reports(celery, 3, api_key="e5444-794", source="fused")
         self.add_reports(celery, 3, api_key=None)
+        self.add_reports(celery, 3, api_key="no-position", set_position=False)
```
jwhitlock: This test confirms the `s3` exporter does not have an issue with `position: None`.

ichnaea/data/tests/test_export.py
```diff
@@ -348,6 +366,18 @@ def test_age(self):
             "wifi": [{"age": -1500}, {"age": 500}],
         }
 
+    def test_null_position(self):
+        assert self.transform(
```
jwhitlock: The fix was in ``InternalTransform`` helper class. This is as close as the code gets to a unit test to confirm the fix.

jwhitlock:
Thanks!

PR #1077, opened 2020-02-19
bug 1616536: add geoip version to __heartbeat__

willkg:
This adds the maxmind db `build_epoch` to the `__heartbeat__` response as geoip version.

To test:

```
curl http://localhost:8000/__heartbeat__
```


willkg:
Thank you!

PR #1078, opened 2020-02-26
bug 1616536: change format of maxmind db version in /__heartbeat__

willkg:
The version is a seconds-since-epoch of the build. This changes it to a
`YYYY-MM-DDTHH:MM:SSZ` format which is what Telemetry uses.

### DIFF
--- a/ichnaea/geoip.py
+++ b/ichnaea/geoip.py
@@ -462,7 +462,7 @@ class GeoIPWrapper(Reader):
         :rtype: string as iso-8601 YYYY-MM-DDTMM:HH:SSZ
         """
         epoch = self.metadata().build_epoch
-        return datetime.datetime.fromtimestamp(epoch).isoformat() + "Z"
+        return datetime.datetime.fromtimestamp(epoch).strftime("%Y-%m-%dT%H:%M:%SZ")
 
     def ping(self):
         """
--- a/ichnaea/geoip.py
+++ b/ichnaea/geoip.py
@@ -4,6 +4,7 @@ Helper functions and classes around GeoIP lookups, based on Maxmind's
 `geoip2 <https://pypi.python.org/pypi/geoip2>`_ Python packages.
 """
 
+import datetime
 import logging
 import time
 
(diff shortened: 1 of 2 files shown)


willkg:
@jwhitlock The thing about this is that since I implemented it as a property on the geoip class, it's easier to set the format to what Telemetry has and use that for the version in heartbeat than it is to have two different formats of the version. Thus I came around to your line of thinking and changed it.

ichnaea/geoip.py
```diff
@@ -458,9 +459,10 @@ def age(self):
     def version(self):
         """
         :returns: The version of the database.
-        :rtype: int
+        :rtype: string as iso-8601 YYYY-MM-DDTMM:HH:SSZ
         """
-        return self.metadata().build_epoch
+        epoch = self.metadata().build_epoch
+        return datetime.datetime.fromtimestamp(epoch).isoformat() + "Z"
```
willkg: The Z thing is gross. I'll look at python-dateutil to see if there's a better way.

jwhitlock: I've had painful experience trying to manually create JS-compatible timestamps with a UTC offset, but I'm having a hard time recalling the details. This is probably close enough, but we should be sure to double-check the generated format in stage.

I was thinking of [specifying seconds](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat), but I think we get that for free when generating from a timestamp:
```suggestion
        return datetime.datetime.fromtimestamp(epoch).isoformat(timespec="seconds") + "Z"
```

willkg: https://stackoverflow.com/questions/19654578/python-utc-datetime-objects-iso-format-doesnt-include-z-zulu-or-zero-offset

That suggests Python by itself won't do the "Z" form for iso-8601 timestamps. So I think we're stuck with "Z".

Also, the tests use a null geoip wrapper, but it's squirrely in that it doesn't actually test the `GeoipWrapper` code at all. I don't want to fix that in this PR, but it means that we have to test it by hand.

I'm going to switch the code to use `strftime` which guarantees the right format and is easier to read.

willkg:
Thank you!

PR #1127, opened 2020-03-06
Enhancement debian package manager tweaks

Rajpratik71:
Major Changes No 1 : debian package manager tweaks

By default, Ubuntu or Debian based "apt" or "apt-get" system installs recommended but not suggested packages . 

By passing "--no-install-recommends" option, the user lets apt-get know not to consider recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog](https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Major Changes No 2 : added packages apt-utils ca-certificates

Because build is 

1.  Slow and in log it is showing because "apt-utils" not installed 

2. to avoid build to exits with error without having certificate

### DIFF
--- a/docker/node/Dockerfile
+++ b/docker/node/Dockerfile
@@ -4,7 +4,9 @@ FROM node:8-slim
 
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get --no-install-recommends install -y \
+    apt-utils \
+    ca-certificates \
     bzip2 \
     git \
     make \
--- a/docker/node/Dockerfile
+++ b/docker/node/Dockerfile
@@ -5,7 +5,6 @@ FROM node:8-slim
 WORKDIR /app
 
 RUN apt-get update && apt-get --no-install-recommends install -y \
-    apt-utils \
     ca-certificates \
     bzip2 \
     git \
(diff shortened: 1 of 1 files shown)


docker/node/Dockerfile
```diff
@@ -4,7 +4,9 @@ FROM node:8-slim
 
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && apt-get --no-install-recommends install -y \
+    apt-utils \
```
jwhitlock: `apt-utils` is used for interactive configuration of packages, and is not needed in this environment. Please remove this line.

docker/node/Dockerfile
```diff
@@ -4,7 +4,9 @@ FROM node:8-slim
...
+    ca-certificates \
```
jwhitlock: I get this message: 

```
ca-certificates is already the newest version (20161130+nmu1+deb9u1).
```

It is already installed in the base image, so we don't need to install it again. Please remove this line.

docker/node/Dockerfile
Rajpratik71: One thing i noticed without "apt-utils" it take longer time to configure the packages and package manager suggest to install "apt-utils"

docker/node/Dockerfile
Rajpratik71: done , may be already in the image . I added because without this wget , curl or git , gives certificate error

PR #1128, opened 2020-03-11
bug 1619789: add API key request instructions for Mozilla projects

willkg:
This adds a separate section for Mozilla projects because we have a different way for them to request API keys.


willkg:
The moco-confidential thing was a new find. I didn't know we could do that before. Yay pre-filled happiness!

PR #1129, opened 2020-03-27
Update to CircleCI 2.1 format

jwhitlock:
Update from version 2.0 to 2.1. The main difference appears to be [pipelines](https://circleci.com/docs/2.0/build-processing/).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 jobs:
   build:
     docker:


jwhitlock:
The tests pass, but the "compiled" config shows version 2.

jwhitlock:
@willkg went through this on socorro: ([pull 5291](https://github.com/mozilla-services/socorro/pull/5291#issuecomment-603511794)):

> Weird. The "configuration" tab seems to say it's still using the 2 config and not 2.1. I verified that the config file is valid, so I'm puzzled.
>
> Then I decided to opt-in to the new user interface because why not throw more problems in the fire to burn and the new user interface differentiates between original config file and the "compiled config file". So maybe it's fine.
>
> This didn't break anything as far as I can tell, so I'm going to land it. self-r+

Same over here. self r+

willkg:
I'm really sorry--I completely forgot to add it to my todo list and then didn't look for it on Monday. I am a bad panda.

PR #1144, opened 2020-04-08
bug 1623470: Drop region tag from metrics

jwhitlock:
As discussed on [bug 1623470](https://bugzilla.mozilla.org/show_bug.cgi?id=1623470), the `region` tag leads to InfluxDB query inefficiencies due to high cardinality.  Thanks to @sciurus for the bulk of this code. My changes are for linting and the metrics docs.

Our plan is to move the `region` tag to the logs, and analyze regions with BigQuery.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -47,6 +47,8 @@ Metric Name                      Type    Tags
 `locate.user`_                   gauge   key, interval
 `queue`_                         gauge   queue
 `region.request`_                counter key, path
+`region.result`_                 counter key, region, accuracy, status, source, fallback_allowed
+`region.user`_                   gauge   key, interval
 `request`_                       counter path, method, status
 `request.timing`_                timer   path, method
 `submit.request`_                counter key, path
@@ -83,6 +85,7 @@ These metrics can help in deciding when to remove a deprecated API.
     and no (``none``) provided API keys.
 
 .. _locate.user:
+.. _region.user:
 .. _submit.user:
 
 API User Metrics
@@ -142,6 +145,7 @@ in it:
     number of valid :term:`stations` for each of the three.
 
 .. _locate.result:
+.. _region.result:
 
 API Result Metrics
 ------------------
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -40,14 +40,14 @@ Metric Name                      Type    Tags
 `locate.fallback.cache`_         counter fallback_name, status
 `locate.fallback.lookup`_        counter fallback_name, status
 `locate.fallback.lookup.timing`_ timer   fallback_name, status
-`locate.query`_                  counter key, region, geoip, blue, cell, wifi
+`locate.query`_                  counter key, geoip, blue, cell, wifi
 `locate.request`_                counter key, path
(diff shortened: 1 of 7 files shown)


ichnaea/api/locate/query.py
```diff
@@ -336,13 +336,9 @@ def collect_metrics(self):
         possible_result = bool(self.expected_accuracy != DataAccuracy.none)
         return allowed and possible_result
 
-    def _emit_region_stat(self, metric, extra_tags):
-        region = self.region
-        if not region:
-            region = "none"
-
+    def _emit_stat(self, metric, extra_tags):
```
jwhitlock, line 339: 👍 nice change, renaming the shared helper function

sciurus:
Heh, I also approve.

Thanks for cleaning this up!

docs/metrics.rst
```diff
@@ -141,7 +140,10 @@ in it:
     one blue, cell and wifi tag get added. The tags depend on the
     number of valid :term:`stations` for each of the three.
 
+*Removed April 2020*: The ``region`` tag
```
willkg, line 143: Sphinx has some additional directives that show nicely in the docs. There's `.. versionadded::`, `.. versionchanged::`, and `.. deprecated::`.

I think I'd use versionchanged here.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded

jwhitlock:
Thanks @willkg, I wasn't going to bug you for a review, but I'm glad you did.

Since we're not doing versioned updates, I used a placeholder for the release tag:

```
.. versionchanged:: 2020-04-unreleased
   Removed the ``region`` tag
```

which renders as:

_Changed in version 2020-04-unreleased_: Removed the ``region`` tag

There will be more of these, so I can add the actual release tag when it ships, and we can adjust as we go forward.

willkg:
I think I've done a version or date--either is fine. I don't think the directive is verifying it's a version.

I used it here with a version and a date:

https://github.com/mozilla/ichnaea/blob/master/docs/api/geosubmit.rst

jwhitlock:
Are you proposing we have a numbered version after 2.2.0? I thought we were using tags like `2020.04.06` instead of incrementing versions now.

willkg:
I'm not proposing anything. I'm saying that the directive doesn't expect a version like you were insinuating in https://github.com/mozilla/ichnaea/pull/1144#issuecomment-611720660 . You could use dates or whatever.

Then I mentioned how we're already using these directives in the docs.

jwhitlock:
OK, I thought you were proposing semantic versioning to get around the issue that we don't know what the release tag will be.

`deprecated` [renders to HTML](https://ichnaea.readthedocs.io/en/latest/api/geosubmit.html) with the prefix text "Deprecated since version ", and `versionchanged` renders to HTML with the prefix text "Changed in version ". I think the release tag (to be finalized later) makes a decent string to complete that sentence. I got the format of [release tags](https://github.com/mozilla/ichnaea/releases) wrong, so it is now:

```
.. versionchanged:: 2020.04.unreleased
   Removed the ``region`` tag
```

This renders as:

_Changed in version 2020.04.unreleased_: Removed the ``region`` tag

Issue #201, opened 2014-05-14
Workaround mis-implementation of geolocate API in FxOS


hannosch:
The code in http://mxr.mozilla.org/mozilla-central/source/dom/system/NetworkGeolocationProvider.js#195 mis-implements the geolocate API. Specifically it sends a `radio` field inside each item in the cellTowers list. This is currently hardcoded to 'gsm', but might be expanded later to the other valid values like wcdma and cdma.

The API mandates sending a `radioType` field in the top-level mapping, as a sibling to the cellTowers entry. It doesn't specify or know about a `radio` field.

Since the wrong implementation is already shipped, we can workaround this on the service end. As no well-behaved client should send the radio field, we can accept and use it, if no radioType field was submitted. A proper radioType field should always win over the radio entries.

PR #1167, opened 2020-05-20
Cleanup metrics tests

jwhitlock:
My intent was to add some additional fixtures and assertions to test and expand the new structured logging ([bug 1632621](https://bugzilla.mozilla.org/show_bug.cgi?id=1632621)). However, I got caught up with making several changes to the existing tests, so I split those into this PR:

* Use the new test methods like ``assert_incr_once`` that were added in [markus](https://github.com/willkg/markus) 2.2.0. This removes at least two lines of code per assertion.
* Add some docstrings to tests. I intended to add these to all the tests I touched, but it took too much time to research git history to understand when and why a test was added, so this work is incomplete (and there are probably more tests to remove or simplify).
* Use ``pytest.mark.parametrize`` to parametrize some tests that make a series of similar calls:
  - ``test_unsupported_methods`` in ``ichnaea/api/locate/tests/base.py`` verifies the response to various unsupported HTTP methods
  - ``test_get_mixed`` in `` ichnaea/api/locate/tests/test_fallback.py`` checks that the cache is not populated when multiple station types are in a query
* Use ``pytest.mark.parametrize`` to combine some tests that are very similar:
  - ``test_cellarea`` in ``ichnaea/api/locate/tests/base.py`` now covers the three ways to enable the ``lacf`` fallback, replacing ``test_cellarea_with_lacf`` and ``test_cellarea_with_different_fallback``.
  - ``test_fallback`` in ``ichnaea/api/locate/tests/base.py`` now covers allowing and disallowing IP-based lookups, replacing ``test_fallback_used_with_geoip``
* Since a ``request`` incr metric is always paired with a ``request.timing`` timing metric, drop most of the tests for ``request.timing``.
* Remove a test ``test_inconsistent_cell_radio_type`` that did less than it appears. It was first written as part of working around a FirefoxOS implementation issue in issue #201. Later, the API was changed, the test file was renamed, tests were re-written to heavily rely on helper functions, and the original work-around is no longer supported. This research is what made me decide to reduce the scope of my test cleanup.

There are a lot of small changes, but most are fairly mechanical, and only test code has changed. Test coverage (lines and branches) is the same after these changes.


ichnaea/api/locate/tests/base.py
```diff
@@ -478,9 +453,14 @@ def test_bad_encoding(self, app, data_queues):
 
 
 class CommonPositionTest(BaseLocateTest):
-    # tests for only the locate_v1 and locate_v2 API's
+    """Tests for the locate_v1 API.
+
+    TODO: These test are split out to support multiple location APIs, but the
+    deprecated APIs were removed in 2019. This could be moved to test_locate_v1.py
+    """
```
willkg, line 460: +1

jwhitlock:
I have a lot of testing instincts that don't make sense with pytest. I still reflexively use ``assert expected == actual``, or as others call it [Yoda conditions](https://wiert.me/2010/05/25/yoda-conditions-from-stackoverflow-new-programming-jargon-you-coined/), because in C ``assert expected = actual`` would fail at compile time if ``expected`` was a constant, but ``assert actual = excepted`` would assign and pass the assertion.

I liked the nose docstrings though 😬 . I can't remember the pain, although I do remember spending too much time writing them, running tests to look at output, re-writing them.

jwhitlock:
I'm added two commits:

* Move the tests from ``CommonPositionTest`` into ``test_locate_v1.py:TestView``
* Replace the ``LocationNotFound`` indirection in tests to use 404 directly. This was needed when `/v1/search` used 200 for location misses.

jwhitlock:
Thanks @willkg!

PR #1169, opened 2020-05-22
bug 1632621: Test structured logging, expand API logging

jwhitlock:
Add a pytest fixture ``logs`` that captures the structured logging entries and makes them available for testing. Add full log entry tests for non-API views, basic successful API views, and some error cases. Add targeted log entry tests for API key validation, and some for location to demonstrate the additional testing capabilities.

This also changes the base API view to allow more log metrics:

* submit APIs now have ``api_type="submit"``
* For geolocate and submit APIs, usage metrics are added:
  - ``api_key_count``: Daily count of API usage by this key
  - ~``api_key_ip_count``~: ~Daily count of API usage by this key and IP~
  - ``api_key_repeat_ip``: True if we've seen this API key / IP combo today (using Redis [HyperLogLog](https://redislabs.com/redis-best-practices/counting/hyperloglog/))~
  - If an API key includes a maximum daily requests limit:
      * ``rate_quota``: (If rate limited) The maximum daily requests for this API key
      * ``rate_remaining``: (If rate limited) The remaining daily requests for this API key, 0 if none
      * ``rate_allowed``: (If rate limited) True if allowed, False if rate limited
* Rate-limiting metrics and the related code are skipped for the region API, to "potentially avoid overhead of Redis connection".
* ``invalid_api_key`` is set when the key in the request is invalid. This can happen when it fails validation (``api_key`` is ``"none"``), or when it is not in the API key table (``api_key`` is ``"invalid"``).
* ``api_key_db_fail=True`` when a database error prevents reading the API key table

### DIFF
--- a/ichnaea/api/locate/tests/base.py
+++ b/ichnaea/api/locate/tests/base.py
@@ -278,7 +278,7 @@ class BaseLocateTest(object):
 class CommonLocateTest(BaseLocateTest):
     """Common tests for geolocate and region APIs."""
 
-    def test_get(self, app, data_queues, metricsmock):
+    def test_get(self, app, data_queues, metricsmock, logs):
         """A GET returns an IP-based location."""
         res = self._call(app, ip=self.test_ip, method="get", status=200)
         self.check_response(data_queues, res, "ok")
@@ -291,16 +291,49 @@ class CommonLocateTest(BaseLocateTest):
             "request.timing", tags=[self.metric_path, "method:get"]
         )
 
-    def test_options(self, app):
+        expected_entry = {
+            # accuracy is low for region API fixture, and medium for geolocate
+            # see bound_model_accuracy and related tests for direct calculation
+            "accuracy": logs.entry["accuracy"],
+            "accuracy_min": "low",
+            "api_key": "test",
+            "api_path": self.metric_path.split(":")[1],
+            "api_type": self.metric_type,
+            "blue": 0,
+            "blue_valid": 0,
+            "cell": 0,
+            "cell_valid": 0,
+            "duration_s": logs.entry["duration_s"],
+            "event": f"GET {self.url} - 200",
+            "fallback_allowed": False,
+            "has_geoip": True,
+            "has_ip": True,
+            "http_method": "GET",
+            "http_path": self.url,
+            "http_status": 200,
+            "log_level": "info",
+            "region": "GB",
+            "result_status": "hit",
+            "source_geoip_accuracy": logs.entry["accuracy"],
+            "source_geoip_accuracy_min": "low",
+            "source_geoip_status": "hit",
+            "wifi": 0,
+            "wifi_valid": 0,
+        }
+        assert logs.entry == expected_entry
+
+    def test_options(self, app, logs):
         """An OPTIONS request works, as required for CORS"""
         res = self._call(app, method="options", status=200)
         assert res.headers["Access-Control-Allow-Origin"] == "*"
         assert res.headers["Access-Control-Max-Age"] == "2592000"
+        assert logs.entry["http_method"] == "OPTIONS"
 
     @pytest.mark.parametrize("method", ("delete", "patch", "put"))
-    def test_unsupported_methods(self, app, method):
+    def test_unsupported_methods(self, app, method, logs):
         """Other HTTP methods are not allowed, and are not logged by app."""
         self._call(app, method=method, status=405)
+        assert logs.entries == []
 
     def test_empty_body(self, app, data_queues, redis):
         """A POST with an empty body returns an IP-based lookup."""
@@ -384,7 +417,7 @@ class CommonLocateTest(BaseLocateTest):
         )
         assert redis.keys("apiuser:*") == []
 
-    def test_invalid_api_key(self, app, data_queues, redis, metricsmock):
+    def test_invalid_api_key(self, app, data_queues, redis, metricsmock, logs):
         """An invalid API key is sanitized to the same as no key."""
         res = self._call(app, api_key="invalid_key", ip=self.test_ip, status=400)
         self.check_response(data_queues, res, "invalid_key")
@@ -392,8 +425,9 @@ class CommonLocateTest(BaseLocateTest):
             self.metric_type + ".request", tags=[self.metric_path, "key:none"]
         )
         assert redis.keys("apiuser:*") == []
+        assert logs.entry["api_key"] == "none"
 
-    def test_unknown_api_key(self, app, data_queues, redis, metricsmock):
+    def test_unknown_api_key(self, app, data_queues, redis, metricsmock, logs):
         """A unknown API key is an error, and increments an "invalid" metric."""
         res = self._call(app, api_key="abcdefg", ip=self.test_ip, status=400)
         self.check_response(data_queues, res, "invalid_key")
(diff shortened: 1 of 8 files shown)


ichnaea/api/views.py
```diff
@@ -140,21 +155,36 @@ def __call__(self):
...
                 skip_check = True
                 self.raven_client.captureException()
-
-        if api_key is not None and api_key.allowed(self.view_type):
+                bind_threadlocal(
+                    api_key=api_key_text,
+                    api_path=self.metric_path,
+                    api_type=self.view_type,
+                    api_key_db_fail=True,
+                )
```
willkg, line 163: This except clause covers any kind of exception--not just a db fail. Is is prudent to restrict this further?

jwhitlock: I hate this style of exception handlers, but it is in the original code. I'll try to determine why this except clause was added and see if there is a more focused exception we can catch.

ichnaea/api/locate/tests/base.py
```diff
@@ -411,6 +454,7 @@ def test_gzip(self, app, data_queues):
         headers = {"Content-Encoding": "gzip"}
         res = self._call(app, body=body, headers=headers, method="post", status=404)
         self.check_response(data_queues, res, "not_found")
+        assert logs.entry["wifi"] == logs.entry["wifi_valid"] == 2
```
willkg, line 457: How does pytest handle this assertion failing? Does it show helpful output?

ichnaea/api/views.py
willkg: There are some cases (*cough*boto*cough*) where it's really hard. I don't know what the case might be here.

ichnaea/api/views.py
```diff
@@ -79,9 +86,17 @@ def log_ip_and_rate_limited(self, valid_key, maxreq):
                 pipe.expire(log_ip_key, 691200)  # 8 days
                 pipe.incr(rate_key, 1)
                 pipe.expire(rate_key, 90000)  # 25 hours
-                _, _, count, _ = pipe.execute()
-                if maxreq and count > maxreq:
-                    should_limit = True
+                ip_count, _, limit_count, _ = pipe.execute()
+                log_params = {
+                    "api_key_count": limit_count,
+                    "api_key_ip_count": ip_count,
```
jwhitlock, line 92: I was wrong about [PFADD](https://redis.io/commands/pfadd). This is written as if it returns the uniqueness count, but instead it returns `1` if it has not seen this value (IP) before and adjusted the cardinality, and `0` if it has seen the value before, and the cardinality is the same.

I'm not sure it is useful to record if this is the first time this API/IP combo was seen today. Either the metric should be dropped, or the name changed.

ichnaea/conftest.py
```diff
@@ -522,3 +525,21 @@ def backoff_sleep_mock():
...
+
+
+class SingleLogCapture(LogCapture):
+    """Capturing processor with helper method for single log entries."""
+
+    @property
+    def entry(self):
+        """Assert there is a single log entry and return it."""
+        assert len(self.entries) == 1
+        return self.entries[0]
```
willkg, line 537: This is interesting. Is it never the case that the code you're testing emits multiple log lines?

willkg: I see other cases so ignore my question.

Mmm... This feels like an assertion on how many entries were emitted in addition to accessing the entry. I think maybe I'd rename this to something different. Maybe `only_entry`? Or `there_can_be_only_one_entry`?

I'm spit-balling. It reads weird to me, but I could get used to it.

ichnaea/api/views.py
jwhitlock: It looks like it was changed to ``Exception`` in 2014 in commit https://github.com/mozilla/ichnaea/commit/25a93524c199cfef379dc063d62de61d8fd5e305.

Looking at https://github.com/PyMySQL/PyMySQL/blob/master/pymysql/err.py and Sentry, I think PyMySQL's  ``DatabaseError`` should catch connection errors and errors due to DB reboots.

jwhitlock: and SQLAlchemy's OperationalError to keep the tests working

ichnaea/api/locate/tests/base.py
jwhitlock: Here's some cases. For the assertion:

```
assert logs.entry["wifi"] == logs.entry["wifi_valid"] == 2
```

``wifi=1``, ``wifi_valid=2``:

```
  File "/app/ichnaea/api/locate/tests/base.py", line 458, in test_gzip
    assert logs.entry["wifi"] == logs.entry["wifi_valid"] == 2
AssertionError: assert 1 == 2
```

``wifi=2``, ``wifi_valid=1``:

```
  File "/app/ichnaea/api/locate/tests/base.py", line 458, in test_gzip
    assert logs.entry["wifi"] == logs.entry["wifi_valid"] == 2
AssertionError: assert 2 == 1
```

Not too clear. I'm going to replace these with just testing ``wifi_valid``:

```
assert logs.entry["wifi_valid"] == 2
```

If ``wifi_valid=1``:

```
  File "/app/ichnaea/api/locate/tests/base.py", line 458, in test_gzip
    assert logs.entry["wifi_valid"] == 2
AssertionError: assert 1 == 2
```

ichnaea/api/views.py
jwhitlock: I'd like to keep this. I'd like a name that is short but self documenting. It would be a True/False value that says if this is the first time we've seen this IP address use this API key today, or if we've already seen this IP address use this API key today.

I did some brainstorming, here's some options sorted by length:

```
first
unique
first_ip
returning_ip
unique_by_ip
good_morning_ip
api_key_repeat_ip
api_key_returning_ip
first_for_key_and_ip
repeat_for_key_and_ip
api_key_first_ip_today
api_key_ip_first_today
api_key_repeat_customer
api_key_repeat_ip_today
first_for_key_and_ip_today
```

I think ``api_key_repeat_ip`` is a good compromise. It keeps it alpha-sorted with the rest of the API metrics, and communicates that it is a repeat IP for this API key. You'd have to dig to see that it starts counting daily, but I think a lot of the log metrics will be day-level anyway. Not perfect, open to inspiration.

ichnaea/conftest.py
jwhitlock: So far, there's exactly one structured log per web request. There are zero logs for static assets and static-like views. I think we lose a lot of the benefits of structured logging if we emit multiple logs per request. So, I like the assertion that there is just one log (most of the time).

This may change when we add structured logging to tasks, which I expect to emit a log per completed task. Or, maybe it won't - this code seems to have avoided ``CELERY_ALWAYS_EAGER`` and the weirdness that causes for Django code. But, if we get into a situation where we have one web log and multiple task logs, does it become ``web_entry``, or ``only_web_entry``? I hate not knowing the future.

I was going to stick with ``entry`` until then, because ``log.entry`` reads better to me than ``log.only_entry``. But because you mentioned it, I've been thinking about it a lot. Then I grepped the code for entry, and got a lot of Sentry code. That will make it even harder to rename in a hypothetical future. So, that's two votes for a longer string.

jwhitlock:
Thanks for the review @willkg! I think your suggestions are real improvements.

ichnaea/api/views.py
willkg: :+1:

ichnaea/api/locate/tests/base.py
willkg: That's good to know! Thank you for testing!

PR #1171, opened 2020-05-28
bug 1635789 - Add SECRET_KEY, response signature

jwhitlock:
Add a ``SECRET_KEY`` configuration setting, used in a similar way to Django's [SECRET_KEY](https://docs.djangoproject.com/en/3.0/ref/settings/#secret-key) as a cryptographic hashing salt. If the web server detects it is not in the development environment, it fails if ``SECRET_KEY`` is set to the non-default value.

Use it to create a response signature, which will be the same for the same combo of response JSON, client IP, API path, and API key. Use [Redis PFADD](https://redis.io/commands/pfadd) to detect unique and repeated responses. Add two new log metrics:
* ``api_repeat_response`` - ``True`` if this response was repeated
* ``api_response_sig`` - The first 8 hex digits of the response signature

### DIFF
--- a/docker/config/test.env
+++ b/docker/config/test.env
@@ -1,6 +1,7 @@
 TESTING=true
 MAPBOX_TOKEN=pk.123456
 ASSET_URL=http://127.0.0.1:9/static/
+SECRET_KEY=default for testing, change in production
 
 # Sets Redis to use "1" db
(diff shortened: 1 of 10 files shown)


jwhitlock:
Thanks @willkg!

PR #1172, opened 2020-05-29
bug 1614130: Update to mapbox.js 3.3.1

jwhitlock:
Update from mapbox.js 3.1.1 to 3.3.1. This includes:

* Using assets included in the NPM package with version 3.2.0. This version also removed the Makefile to construct the assets yourself.
* Switch from the deprecated `a.tiles.mapbox.com` CDN to `api.mapbox.com` for the base tiles, which required some CSP updates
* Upgrade jQuery 3.5.0 to 3.5.1, a minor release to fix a regression

I tested this locally by setting the needed variables in `my.env`:

```bash
MAPBOX_TOKEN=pk.my_personal_dev_token  # Signed up for a free account months ago
ASSET_URL=https://d3r3tk5171bc5t.cloudfront.net
```

I then ran the site, loaded the web developer console, and looked at the two maps pages:
* http://localhost:8000/ - Has a small full world map
* http://localhost:8000/map - Interactive world map


ichnaea/content/views.py
```diff
@@ -31,16 +31,18 @@
 
 CSP_BASE = "'self'"
 CSP_POLICY = """\
-default-src 'self' *.tiles.mapbox.com;
+default-src 'self';
+connect-src {base} api.mapbox.com;
```
jwhitlock, line 35: The URL for the old tiles was `https://a.tiles.mapbox.com/v4`, this changes to the new URL `api.mapbox.com`. There is [CSP guidance](https://docs.mapbox.com/mapbox-gl-js/overview/#csp-directives) for the GL API (vector tiles), but not the static API. It does mention ``https://*.tiles.mapbox.com``, but my testing didn't show any calls to that URL, or ``https://events.mapbox.com``.

I did see errors about the GA code trying to connect, such as this from Safari:

```
Refused to connect to https://www.google-analytics.com/j/collect?<omitted> because it does not appear in the connect-src directive of the Content Security Policy.
```

I see similar errors in production with Safari, so I suspect Google Analytics has been broken for a long time. I don't think we need to analyze web site visitors for MLS, so I filed [bug 1642422](https://bugzilla.mozilla.org/show_bug.cgi?id=1642422) to remove Google Analytics.

ichnaea/content/views.py
```diff
@@ -31,16 +31,18 @@
...
+img-src {base} {tiles} *.google-analytics.com api.mapbox.com data:;
 script-src {base} *.google-analytics.com data: 'unsafe-eval';
 style-src {base};
 """
 CSP_POLICY = CSP_POLICY.replace("\n", " ").strip()
 TILES_PATTERN = "{z}/{x}/{y}.png"
 HOMEPAGE_MAP_IMAGE = (
-    "https://a.tiles.mapbox.com/v4/mapbox.dark" "/0/0/0@2x.png?access_token={token}"
+    "https://api.mapbox.com/styles/v1/mapbox/dark-v10/tiles"
+    "/256/0/0/0@2x?access_token={token}"
```
jwhitlock, line 45: This is the "whole earth" tile seen on the homepage, and it continues to work.

docker/node/Dockerfile
```diff
@@ -6,7 +6,6 @@ WORKDIR /app
 
 RUN apt-get update && apt-get --no-install-recommends install -y \
     bzip2 \
-    git \
```
jwhitlock, line 9: ``git`` may have been needed by earlier versions of the build process, but does not appear to be needed now. ``bzip2`` is still needed to unpack ``phantoms``, and ``make`` is used to create the bundled assets.

PR #1219, opened 2020-06-10
bug 1635789: Remove PFADD log metrics

jwhitlock:
The metrics ``api_key_repeat_ip`` and ``api_repeat_response`` are based on the
return from [PFADD](https://redis.io/commands/pfadd), used to estimate the number of daily unique IP addresses and responses. After further research, this does not signify if a value is a repeat, but if the estimated cardinality has changed. This is less and less frequent as the cardinality increases, exceeding 99% at 100 million items. This is not a useful metric to track.

The cardinality of the IP addresses is useful, so the related ``PFADD`` call is retained, without capturing the return. The [PFCOUNT](https://redis.io/commands/pfcount) return, accurate to about 1%, is still sent [as an operational metric](https://github.com/mozilla/ichnaea/blob/b3f523a8bc6bc20f310317b7cd19c45c460edd15/ichnaea/data/monitor.py#L56-L62).

The cardinality of unique responses is not useful, so the related `PFADD` call is removed. The signature may still be useful in BigQuery analysis (such as on [bug 1635789](https://bugzilla.mozilla.org/show_bug.cgi?id=1635789)), so we keep it for now as a log metric.

### DIFF
--- a/ichnaea/api/locate/tests/base.py
+++ b/ichnaea/api/locate/tests/base.py
@@ -325,8 +325,6 @@ class CommonLocateTest(BaseLocateTest):
         }
         if self.metric_type == "locate":
             expected_entry["api_key_count"] = 1
-            expected_entry["api_key_repeat_ip"] = False
-            expected_entry["api_repeat_response"] = False
             expected_entry["api_response_sig"] = log["api_response_sig"]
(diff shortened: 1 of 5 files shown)


jwhitlock:
Thank you!

PR #1220, opened 2020-06-25
bug 1647364: Switch default branch from master to main

jwhitlock:
Update the configuration and documentation for today's change of making the new `main` branch the default branch (see [this blog post](https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx), as well as some other updates:

* In CircleCI, take special actions like uploading docker images on the `main` rather than `master` branch
* Update `bin/release.py` to latest version that supports different default branches, plus a line length change
* Update documentation references and code comments to the `main` branch of Ichnaea
* Bump copyright to 2020
* Fix some URLs and surrounding docs that failed `make linkcheck` validation

Addition, non-code changes are also needed, see [bug 1647364](https://bugzilla.mozilla.org/show_bug.cgi?id=1647364) for tracking.

PR #1221, opened 2020-06-26
bug 1647364: Add master_doc to config

jwhitlock:
readthedocs.org builds with Sphinx v1.8.5, and `master_doc` defaults to `'contents'` before Sphinx v2.0. This sets `master_doc='index'`, the Sphinx v2.0 and later default.

Alternatively, we could refactor the documentation dependencies into a new file, so that RTD gets the same version of Sphinx as the development environment.

Follow-on work for [bug 1647364](https://bugzilla.mozilla.org/show_bug.cgi?id=1647364) and PR #1220.


jwhitlock:
After some discussion, we should instead install the same version of Sphinx in RTD that we use in development.

PR #1222, opened 2020-06-26
bug 1647364: Install Sphinx 3 in RTD

jwhitlock:
Split documentation dependencies so that they can be installed in ReadTheDocs. This will allow us to use Sphinx 3.0.4 in both dev and in RTD.

The change in PR #1220 broken documentation builds, because RTD uses Sphinx 1.8.5 by default, and I was relying on a change in 2.0. I also followed the [FAQ](https://docs.readthedocs.io/en/stable/faq.html#how-do-i-change-behavior-when-building-with-read-the-docs) to use the default RTD theme in RTD.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import os
-import sphinx_rtd_theme
 import sys
 from unittest import mock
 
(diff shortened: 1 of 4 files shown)


jwhitlock:
Thanks! I've updated settings on RTD, maybe it will work the first time? 🤞

jwhitlock:
Nope! Next PR incoming...

PR #1223, opened 2020-06-26
bug 1647364: Use the new refactored requirements/docs.txt

jwhitlock:
I missed that we already had a solution that installed requirements. This builds on PR #1222, to switch the ReadTheDocs config to use the new refactored ``requirements/docs.txt``.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -10,6 +10,6 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.6
+  version: 3.8
   install:
-    - requirements: docs/requirements.txt
+    - requirements: requirements/docs.txt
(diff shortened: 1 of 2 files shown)


.readthedocs.yml
```diff
@@ -10,6 +10,6 @@ formats: all
 
 # Optionally set the version of Python and requirements required to build your docs
 python:
-  version: 3.6
+  version: 3.8
```
jwhitlock, line 13: The [build.image](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-image) docs say that the `latest` builder (the default) has 3.8, so this syncs with the rest of the project.

jwhitlock:
self r+, non-prod code and RTD will tell me what I did wrong...

PR #1287, opened 2020-08-03
Update dependencies, August 2020, round 3: More production dependencies

jwhitlock:
Update production dependencies that took a little more work. This covers:

* Bump boto3 from 1.14.16 to 1.14.33 (from PR #1272) _(upgrades cleanly with botocore updated)_
* Bump geoip2 from 3.0.0 to 4.0.2 (from PR #1270) _(Python 3.6 required, new async web service client with additional dependencies)_


jwhitlock:
Tested locally with:

* ``make build lint docs test``
* ``make run``, poke around, http://localhost:8000/__heartbeat__ shows that GeoIP could be read
* ``make runcelery``, stare at scheduled tasks

PR #1288, opened 2020-08-03
Move urllib3[security] to requirements/shared.txt

jwhitlock:
The new pip dependency resolver, planned for the 20.3 release, no longer accepts constraints with extras, so we need to move ``urllib3[security]`` out of ``contraints.txt``. It is also needed by Sphinx, so it moves to ``shared.txt``.

Before this PR, with pip 20.2, a warning is printed before packages start installing:

```
DEPRECATION: Constraints are only allowed to take the form of a package name
and a version specifier. Other forms were originally permitted as an accident of
the implementation, but were undocumented. The new implementation of the
resolver no longer supports these forms. A possible replacement is replacing the
constraint with a requirement.. 
You can find discussion regarding this at https://github.com/pypa/pip/issues/8210.
```

I was unable to install our requirements locally with ``pip install -r requirements/default.txt --use-feature=2020-resolver``, getting this error:

```
    chardet<4,>=3.0.2 from https://files.pythonhosted.org/packages/bc/a9/01ffebfb562e4274b6487b4bb1ddec7ca55ec7510b22e4c51f14098443b8/chardet-3.0.4-py2.py3-none-any.whl#sha256=fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691 (from requests[security]==2.24.0->-r requirements/shared.txt (line 9))
```

This may be a legit bug in the 2020 resolver. I'll tell the pip project.


jwhitlock:
self +r, installs the same but without a warning on pip 20.2.

PR #1314, opened 2020-09-08
Tweaks based on setting up local development

djmitche:
* Avoid an ugly traceback in `make setup`
* UID/GID is required on Linux (although the default of 1001 will probably work for most people, if they're the only user on their laptop)


jwhitlock (approved):
Thanks @djmitche, the changes work for me:

* I'm on macOS so I can't test ``id`` and setting the UID / GID, but yes, Linux users *must* do this or they will have a bad time
* The new ``drop_db`` function and ``drop_database`` script work for me. ``make setup`` worked when there was no location database, and then again with a database. ``make setup`` continues to fail if I forget to run ``make runservices`` first...
* Thanks for confirming ``make update_vendored`` works, and updating MaxMind test databases to ``2020-08-27T15:50:01Z``.

djmitche:
> Thanks for confirming make update_vendored works, and updating MaxMind test databases to 2020-08-27T15:50:01Z.

I somehow totally missed that this made changes in version control!

PR #1316, opened 2020-09-08
run lint over specific directories, fix black issues

djmitche:
I keep my Python virtualenvs in the root of the repository, so a bare `pyflakes8` or `black` will try to lint everything I have installd -- no fun.

It seems like black skips a few files by default, too, that are now covered, and are updated in this PR (just changing quote symbols).


jwhitlock (approved):
Thanks @djmitche!

PR #1322, opened 2020-09-18
1632621: Rewrite metrics doc for structured logs

jwhitlock:
Rewrite the metrics document and expand it to cover structured logs:

* Describe the difference between statsd metrics and structured logging
* Create an Implementation section, to describe our implementation, the issues around high cardinality, and that this is all in progress
* Instead of the narrative flow, split up metrics by web app, task app, and datamaps, with some secondary sections as needed, and a new section for web app structured logs.
* Split up metrics to a section per metric, with a new style for documenting tags, and with cross-references to structure logs
* Reorganize backend and pipeline sections, and add more context. Some of this is duplicated on the data flows page, and could be merged.

All of this makes the diff view fairly useless for reviewing this change (except for the metrics table, which gained ``region.query``). A reviewer may want to cross-reference to the [current metrics document](https://ichnaea.readthedocs.io/en/latest/metrics.html) instead.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -12,7 +12,7 @@ Ichnaea provides two classes of runtime data:
 Structured logs were added in 2020, and the migration of data from metrics to
 logs is not complete. For more information, see the Implementation_ section.
 
-Metrics are emitted by the web / API application, the async task application,
+Metrics are emitted by the web / API application, the backend task application,
 and the datamaps script:
 
 ================================ ======== ======= =======================================================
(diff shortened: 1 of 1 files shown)


jwhitlock:
Feel free to read and review, but I think this will be a self-review. I will wait a day and then re-read with fresh eyes.

jwhitlock:
I made one big change in the second round, adding an "App" column to the metrics table. I made several smaller changes, re-writing some sentences, trying to get more consistent about periods at the end of bullet points, etc.

I think there is more to do, but that can be tackled as we add more structured logging and re-write the [data flows doc](https://ichnaea.readthedocs.io/en/latest/dataflows.html) to add some diagrams.

PR #1341, opened 2020-10-01
Update test.env

uvinduperera:


jwhitlock:
Thanks @uvinduperera, but I do not think we can accept this change. This now shows up in the code as a quoted string:

```shell
$ make testshell
app@173c1a9b5b28:/app$ env | grep SECRET_KEY
app@173c1a9b5b28:/app$ python
>>> from ichnaea.conf import check_config, settings
>>> settings.config.options.options['secret_key'].default
'default for development, change in production'
>>> settings('secret_key')
'"default for testing, change in production"'
>>> settings.config.options.options['secret_key'].default == settings('secret_key')
False
```

This check is used in ``check_config`` to ensure that the ``SECRET_KEY`` is changed in a production environment:

https://github.com/mozilla/ichnaea/blob/43e51ee7aa347db82e05e9c28f19fca675e76325/ichnaea/conf.py#L150-L172

If we're adding the quotes in ``test.env``, we'd need to also add then in ``ichnaea/conf.py``, and probably in a test.

Was the lack of quotes causing an issue?

PR #1345, opened 2020-10-01
Update package.json

uvinduperera:


jwhitlock (approved):
Thanks @uvinduperera! We're currently installing mapbox.js 3.3.1, so this doesn't change what ships.

I did see this warning:

```
npm WARN notsup Unsupported engine for mapbox.js@3.3.1: wanted: {"node":">=10"} (current: {"node":"8.17.0","npm":"6.13.4"})
npm WARN notsup Not compatible with your version of node/npm: mapbox.js@3.3.1
```

It is not caused by this change, we get a similar warning before this PR. Updating the node engine is reported in [bug 1642372](https://bugzilla.mozilla.org/show_bug.cgi?id=1642372).

PR #1363, opened 2020-10-05
Switch to dependabot v2

jwhitlock:
Switch from the dependabot-preview spec to the current version.

Based on https://github.com/mozilla-services/antenna/pull/635. @willkg does this look correct?


willkg (approved):
This is what I did with the socorro/tecken things and it worked super. r+!

PR #1369, opened 2020-10-05
Login to docker before building images

jwhitlock:
Starting [November 2020](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/), Dockerhub will limit unauthenticated docker pulls by IP address. Log in before building images, which may pull from DockerHub, using [a setup](https://circleci.com/docs/2.0/private-images/) suggested by CircleCI, and building on login before pushing tags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,6 +47,15 @@ jobs:
                docker-compose --version
 
         - run:
+            name: Login to Dockerhub
+            command: |
+                if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
+                  echo "Skipping Login to Dockerhub, credentials not available."
+                else
+                  echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
+                fi
+
+        - run:
             name: Build Docker images
             working_directory: /ichnaea
             command: |
@@ -100,6 +109,8 @@ jobs:
                   docker tag "local/ichnaea_app" "mozilla/location:latest"
                   retry docker push "mozilla/location:latest"
                 fi
+              else
+                echo "Skipping Push to Dockerhub, not main or tag."
               fi
 
 workflows:


jwhitlock:
We use dockerhub in three ways:

* The CircleCI builder is docker image [mozilla/cidockerbases:docker-latest](https://hub.docker.com/r/mozilla/cidockerbases/)
* The ichnaea images are based on [python:3.8.6-slim](https://hub.docker.com/_/python) and [node:8-slim](https://hub.docker.com/_/node)
* The ichnaea tagged images are uploaded to [mozilla/location](https://hub.docker.com/r/mozilla/location)

Previously, we only logged into Dockerhub for the last case, so we've restricted passing environment variables (including the docker credentials) to branches on the main repo at https://github.com/mozilla/ichnaea. This way, only developers authorized to push to the main repo can create branches that can interact with Docker. Other PRs from a fork, like this one, don't get those credentials.

Some options:
1. Do nothing:
  * Wait and see if any of this actually a problem. We can merge this PR once we have a feel for how necessary it is.
  * Wait and see if CircleCI works something out with Dockerhub, and switch to any recommended solution like a CircleCI-hosted docker mirror.
2. Use this solution:
  * Switch from a fork-and-pull model to a shared repository model. This would mean that external contributors are more likely to hit the docker anonymous IP limits. Dependabot makes branches on this repo, so the monthly updates would not be affected.
  * Use other credentials that don't have push access for forked projects. These would either be hard-coded into ``.circleci.yml`` or maybe there would be a way to add them selectively. However, there is still an increased chance for mischief from malicious PRs.
3. Migrate away from Dockerhub
  * Move Mozilla's base image to another docker image repository
  * Setup a mirror, or use one like https://cloud.google.com/container-registry/docs/pulling-cached-images
  * Start publishing our location image elsewhere

I'm inclined to wait a bit to see what others figure out, then explore migrating away from Dockerhub.

.circleci/config.yml
```diff
@@ -46,6 +49,11 @@ jobs:
             name: Build Docker images
             working_directory: /ichnaea
             command: |
+                if [ "${DOCKER_USER}" == "" ] || [ "${DOCKER_PASS}" == "" ]; then
+                  echo "Skipping Login to Dockerhub, credentials not available."
+                else
+                  docker login --username "${DOCKER_USER}" --password "${DOCKER_PASS}"
```
sciurus, line 55: Is doing this still necessary, or is it redundant now that you have the `docker: auth:` block ?

jwhitlock: It's a good question. I'm not sure what ``docker login`` does under the hood in CircleCI. Here's the [docker login](https://docs.docker.com/engine/reference/commandline/login/) docs, there are a few ways to store credentials.

When I use it locally with personal credentials, ``~/.docker/config.json`` says I'm using ``"credsStore": "desktop"``. This suggests to me that Docker Desktop is storing credentials in memory. When I run ``docker login`` twice, it says "Authenticating with existing credentials..."

The [CircleCI config reference](https://circleci.com/docs/2.0/configuration-reference/#docker) doesn't say how things are done. I could imagine:

1. You just login once, and the rest of the docker commands use the same authenticated session
2. You need to login at the level of fetching the docker image, and again for build steps using the docker image
3. You need to login to fetch the image, and again for _each_ step that runs a docker command.

Some experimentation may be needed to figure out which applies. I'd be quite happy if one login only was needed. I believe I can test this tomorrow, to see if I can tag and push a release with just the ``docker: auth`` block.

I tried to use ``--password`` instead of ``password-stdin`` here, but when I try it locally, I get a warning:

```
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
```

So, I'll at least want to change that.

jwhitlock: I tried just the ``docker: auth`` block in #1372. It worked as a PR, but I wasn't authenticated for the push when merging to master. I think we need at least one ``docker login`` in the build steps as well.

jwhitlock:
Force-push to rebase on master, especially to get PR #1372 integrated. Create an explicit "Login to Dockerhub" step, and do it before making the images.

I think further changes may be needed. The [discussion](https://discuss.circleci.com/t/authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567/13) makes me think more changes are coming to CircleCI's implementation. However, our Dockerhub push is currently broken, and this should fix it. Or, if it doesn't, then we learn that ``docker login`` needs to be in the same step as the other docker commands.

Issue #1417, opened 2020-11-02
Update to Python 3.9


jwhitlock:
Python 3.9 is released. The Dockerfile update is PR #1410, and fails when installing ``scipy``.

The final error line is:

```
numpy.distutils.system_info.NotFoundError: No lapack/blas resources found.
```

On a scipy issue, a maintainer suggests 3.9 wheels are coming and more system packages are needed (specifically ``libatlas-base-dev liblapack-dev gfortran``:

https://github.com/scipy/scipy/blob/master/.github/workflows/bleedingedge.yml#L32

There may be a sequence of similar changes like this.

jwhitlock:
The ``scipy`` and ``numpy`` wheels are now available, so we're unblocked on a Python 3 release.

I'd like to read the [release notes](https://docs.python.org/3.9/whatsnew/3.9.html) and get the December updates in production first.s

jwhitlock:
I ran ``pytest -Wd ichnaea`` to look for Python 3.9 deprecations. The only warnings are the ``ResourceWarning``s tracked in  issue #1393.

One change for us might be the new [zoneinfo](https://docs.python.org/3.9/whatsnew/3.9.html#zoneinfo) module. Some of our requirements load [pytz](https://pythonhosted.org/pytz/), and we use it directly or UTC. With the Olsen timezone database built in, we could drop that as a top-level requirement.

The new dictionary merge (``|``) and update (``|=``) operators could be useful, as well as ``str.removeprefix()`` and ``str.removesuffix()``.

jwhitlock:
I'm tracking the `zoneinfo` update in issue #1440

PR #1422, opened 2020-11-06
issue #1417: Add libs for scipy and Py 3.9

jwhitlock:
``scipy`` fails to build under Python 3.9, because wheels are not yet ready and the source build can't load lapack/blas resources:

```
numpy.distutils.system_info.NotFoundError: No lapack/blas resources found.
```

This adds the ``gfortran``, ``liblapack-dev``, and ``libatlas-base-dev`` libraries, needed to build from source, as seen in the [bleeding edge](https://github.com/scipy/scipy/blob/master/.github/workflows/bleedingedge.yml#L32) build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,10 +36,13 @@ RUN apt-key add /tmp/mysql_pubkey.asc && \
 RUN apt-get update && \
     apt-get -y install --no-install-recommends \
     file \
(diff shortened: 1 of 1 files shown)


jwhitlock:
My local tests would be the same as the CI tests, so trusting CI.

Issue #1398, opened 2020-10-27
Observation queues exceed maximum size targets


jwhitlock:
*Originally filed on 2020-10-07 as [bug 1669825](https://bugzilla.mozilla.org/show_bug.cgi?id=1669825):*

After [bug 1613493](https://bugzilla.mozilla.org/show_bug.cgi?id=1613493) (*"increase processing sample rate in prod", closed, confidential*), MLS observation queues were stable, by setting the sample rate of the largest geolocate API user to 35%. Since then, the queues have occasionally exceeded the target of no more than 10 million observations to process. The standard rate has been lowered, now at 10%.

Some of this may be due to increased usage and more data per day. Some may also be due to the growing database causing operations to slow down (issue #845 may address this).

One issue is that processing stops for 10-15 minutes during a production deploy, while the observations continue to accumulate. The processing system has difficulty catching up, even during the overnight slow periods.

The solution has been:

1. Reduce the sample rate for the large API user to 1%
2. Wait for the queues to reduce to a low level (10K - 100K each), 1 to 6 hours
3. Return the sample rate to the original value

One way to automate this would be to add a global sample rate and a maximum queue size. As the total observation backlog approaches the maximum, the global rate can be reduced from 100%, slowing the rate of incoming observations. When the backlog is back under control, the global rate can grow back to 100%, if the async workers can handle the incoming data.

*Updated on 2020-10-08:*

The queues briefly cleared around midnight UTC, but have been growing steadily since 8 AM.

The deploy included a recent update from celery 4.4.7 to 5.0.0 (PR #1362)]. The [major version bump](https://docs.celeryproject.org/en/stable/whatsnew-5.0.html) was due to removing Python 2 support, but there are other changes.

There were also changes to the greenlet library (PR #1359) to support contextvars, and a coordinating change to the gevent library.

One of the changes is in the Redis results backend. Group results are now in invocation order by default. We use Redis as the results backend, but don't use chained tasks or group results in general, so this should not affect our tasks.

ckolos noticed that the [task workers CPU usage is lower](https://earthangel-b40313e5.influxcloud.net/d/ULUTsaSZx/location-prod-infra-dash?orgId=1&from=1600838543239&to=1602169907605&var-DB=location-prod-master-20191126&var-autoscale_group_name=) (*Mozilla Graphana instance*). Previously, the peak CPU was around 65% on weekdays, dropping to 45% at night. Now tasks are more consistently at 25% - 33% at all hours.

[...]

PR #1424, opened 2020-11-17
issue 1398: Remove countdown from station tasks

jwhitlock:
Remove the countdown parameter so that station update tasks (WiFi, etc.) are executed as soon as possible, rather than spaced out with a 5 second delay. This should have an impact on issue #1398.


jwhitlock:
Merging for integration tests on staging

PR #1427, opened 2020-11-21
Switch to pip-compile

### LINKED ISSUE #1407: Use the pip 2020 resolver (default in 20.3.x)
Install python libraries with pip 20.2 and the 2020-resolver planned for the 20.3 release. See the [changes to the pip dependency resolver in 20.2 (2020)](https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-2-2020)

This fails for me when installing in the docker image, and when installing on macOS in a virtualenv. It succeeds when the default resolver in pip 20.2 is used.

The error with ``--use-feature=2020-resolver`` is:

```
    idna<3,>=2.5 from https://files.pythonhosted.org/packages/a2/38/928ddce2273eaa564f6f50de919327bf3a00f091b5baba8dfa9460f3a8a8/idna-2.10-py2.py3-none-any.whl#sha256=b97d804b1e9b523befed77c48dacec60e6dcb0b5391d57af6a65a312a90648c0 (from requests[security]==2.24.0->-r requirements/shared.txt (line 9))
```

Previous work was in PR #1289. Some highlights:

* https://github.com/jwhitlock/pip-resolver-demo demonstrates the issue
* The new resolver was randomly applied and broke documentation, reported and fixed
* A fix was discussed and shipped in pip 20.2.4, but didn't fix our issue
* The pip 20.2.4 issue was discussed in https://github.com/pypa/pip/issues/9020. It is likely that the the next action is to document this as a known regression, and to discourage using hashes with constraints.

My next attempt will be to follow @willkg's strategy in https://github.com/mozilla-services/socorro/pull/5595 and switch to ``pip-compile`` instead. I also need to adjust my advocacy for [constraints as requirements of requirements](https://stackoverflow.com/a/36848206/10612). They are a lot less useful with the new resolver.

jwhitlock:
``pip-compile``, part of [pip-tools](https://github.com/jazzband/pip-tools/), provides a new process for maintaining Python dependencies. The top-level dependencies are defined in a ``requirements.in`` file, and ``pip-compile`` will determine the dependencies-of-dependencies and calculate hashes, writing them to ``requirements.txt``.

This places the dependencies in four files, ``shared.in``, ``docs.in``, ``prod.in``, and ``dev.in``. After running ``make update-reqs``, these generate ``docs.txt`` (for ReadTheDocs) and ``dev.txt`` (for the Docker image).

Other changes:
* Adds some documentation that lives in the ``requirements`` folder.
* Rewords some development documentation. This was where I was going to put the requirements docs, but these seemed to be the "new developer" docs, and the requirements doc was long enough for its own page. Still, it seemed awkward to me
* Removes ``setup.py``, which we haven't used in a while.

**Update 1**: Changed from ``make reqs-regen`` to ``make update-requirements``, removed CI check, added new docs page.

**Update 2**: ``make update-reqs`` is shorter, doesn't break formatting of ``make help``. Also, added in ``--use-feature=2020-resolver``, so now this fixes the issue.

Fixes #1407


jwhitlock (commented):
Works locally and in CI. I used ``pip freeze`` to compare the installed requirements before and after. To keep requirements from changing, I added the section ``#TODO: Unpin these with the December update``.

I'm merging now, so I can see how Dependabot works with this workflow.

Issue #1428, opened 2020-12-01
Dependabot unable to update project


jwhitlock:
After PR #1427, Dependabot had an error that it was unable to update the project. The specific error was that it couldn't find a match for ``amqp==5.0.1``, and listed earlier versions before Python 3 was required.

I see a few things to try:
* Press the "try again" button.
* Add ``setup.py`` back, which specifies Python 3.6
* Restrict Dependabot to the ``requirements`` directory

jwhitlock:
Here's the new error from the retry:

> Dependabot can't resolve your Python dependency files.
>
> As a result, Dependabot couldn't update your dependencies.
>
> The error Dependabot encountered was:
>
> Could not find a version that matches aio<redacted>
[...]

PR #1430, opened 2020-12-03
Merge to a single requirements.in / .txt

jwhitlock:
Dependabot is not able to update our dependencies (issue #1428). I tried to manually update them using ``pip-compile --upgrade``, and found some difficulties:

* ``pip-compile --upgrade`` will not upgrade a package that is pinned in the ``*.in`` file. I tried removing these version pins
* An upgrade of one requirement file may work, but the next would fail. For example, ``docs.in`` updated ``requests==2.25.0`` as a secondary dependency, then ``prod.in`` failed to update because it specified ``requests[security]==2.24.0``.

After a few attempts, I'm going to the simpler setup of:

* ``requirements.in`` - The primary package requirements for both production and development. For now, they are still pinned. I'll retry Dependabot and see if it updates them, otherwise I'll remove them so I can use ``pip-compile --upgrade`` manually for December.
* ``requirements.txt`` - All the prod/dev requirements. It has the same contents as ``requirements/dev.txt`` did, except for different comments from ``pip-compile``.
* ``requirements-docs.txt`` - A simple requirements file (no pins, no hashes) to get the documents building in ReadTheDocs.

I've also removed ``botocore`` from the first-level requirements. Even though it is directly imported by the production code, it is automatically included by ``boto3``, and the mutual update of those two packages causes a headache every month.


jwhitlock:
CI is happy. Let's see if ReadTheDocs and Dependabot are happy as well.

Issue #1421, opened 2020-11-06
outdated library: mobile-codes causes ignoring some data


szjozsef:
The mobile-codes library is marked unmaintained and is really out of date, is used internally to validate the data (at least during import).
The library code itself is ok, but the included json databases are out of date.
I managed to update the 2 relevant files, 
`
mobile-codes/mobile_codes/json/countries.json
mobile-codes/mobile_codes/json/mnc_operators.json
`
and mapped them into the container via :
`
-v "/host_folder/mnc_operators.json:/usr/local/lib/python3.8/site-packages/mobile_codes/json/mnc_operators.json" \
-v "/host_folder/countries.json:/usr/local/lib/python3.8/site-packages/mobile_codes/json/countries.json" \
`
Also the library's github repository do contain a parse.py script which should update this files, but that script is also out of date.
For the moment I published the updated files in a forked repository: `https://github.com/szjozsef/mobile-codes`
Maybe ichnaea should update internally this files from time to time (?), or at least document somewhere that if the files get outdated, data loss can occur.
I have updated the parse.py but because I have limited python knowledge only succeed to obtain the updated files with a combination of shell commands and the python script. If somebody with good python knowledge are willing to update the parse.py I can share my scripts with her/him.

jwhitlock:
Thank you for your work on this problem, @szjozsef!

I see a couple of options to update this data:

1. Reach out to andymckay (an ex-Mozilla employee) and see if he is willing to transfer the project to Mozilla, so we can maintain it
2. Bring the code and data into this project, so we can maintain it

I searched for users of this library on GitHub:

https://github.com/search?q=path%3Arequirements+mobile-codes

The results were:

* This project
* https://github.com/mozilla/solitude - Archived. Solitude was the Payment Server for processing marketplace and addons payments.
* https://github.com/mozilla/webpay - Archived. Webpay is an implementation of the WebPaymentProvider spec. It hosts the payment flow inside ``navigator.mozPay()`` when making app purchases or in-app payments on Firefox OS.

I couldn't find any non-Mozilla projects or active projects that use this library. That makes me want to absorb the project into this one rather than keep it as an independent project.

I then looked at the forks:
* https://github.com/drstevens/mobile-codes - 1 change Nov 2015, switched the Wikipedia page, made some other script changes, regenerated output
* https://github.com/nToggle/mobile-codes - Same as drstevens
* https://github.com/impratikjaiswal/mobile-codes - No changes
[...]

PR #1442, opened 2020-12-11
Update mobile-codes and data from szjozsef's branch

jwhitlock:
For issue #1421, update the recently adopted ``mobile-codes`` (PR #1444) with @szjozsef's changes in [szjozsef/mobile-codes](https://github.com/szjozsef/mobile-codes).

Follow-on issues:
* #1445 - Use JSON name/value pairs
* #1446 - Standardize data mapping


jwhitlock:
I've got the new ``parse.py`` running, but the files it outputs break current tests. The test failures:

```
_______________________________________________ TestCountries.test_mcc_multiple_codes _______________________________________________
  File "/app/mobile_codes/tests.py", line 14, in test_mcc_multiple_codes
    assert len(countries) == 1
AssertionError: assert 2 == 1
 +  where 2 = len([Country(name='Puerto Rico', alpha2='PR', alpha3='PRI', numeric='630', mcc=['310', '313', '330']), Country(name='United States of America', alpha2='US', alpha3='USA', numeric='840', mcc=['310', '311', '312', '313', '314', '315', '316'])])
_________________________________________ TestCountries.test_countries_match_mnc_operators __________________________________________
  File "/app/mobile_codes/tests.py", line 86, in test_countries_match_mnc_operators
    assert countries_mccs - operator_mccs == set()
AssertionError: assert {'312'} == set()
  Extra items in the left set:
  '312'
  Use -v to get the full diff
____________________________________________ TestCountriesSpecialCases.test_puerto_rico _____________________________________________
  File "/app/mobile_codes/tests.py", line 117, in test_puerto_rico
    assert country.mcc == ["310", "330"]
AssertionError: assert ['310', '313', '330'] == ['310', '330']
  At index 1 diff: '313' != '330'
  Left contains one more item: '330'
  Use -v to get the full diff
```

These could be legitimate changes - it has been a few years and there could be new codes. Or, something may be wrong with the import.

The previous code worked on a manually downloaded copy of the Wikipedia page. The new code takes care of downloading the page as well. I'd like to split out the downloading into its own method, so that I can work on the parsing without network slowdowns, and maybe pinpoint the Wikipedia revision of the source pages.

There's also a number of string replacements for country names, such as changing "United Kingdom" to "United Kingdom of Great Britain and Northern Ireland". I'd like to figure out a way to list these string fixups outside of Python code, since it seems like something that will change often.

The merge code has issues around nested lists, like:

```diff
-    ["French Guiana", "GF", "GUF", "254", "340"],
+    ["French Guiana", "GF", "GUF", "254", ["340", "742"]],
```

This is probably just an ``extend`` rather than ``append``. There's also some ``null`` in the lists, and I'm not sure what that's about.

Still, we're a big step closer to updated mobile codes!

szjozsef:
Please see the comment: https://github.com/mozilla/ichnaea/issues/1421#issuecomment-744457048
The MCC code in the countries.json can cotain:
```
    None (no MCC code)   --> This corresponds to the `null` values
    a string (where a country has one code) --> example: "340"
    a tuple of strings (where a country has more than one code) --> example: ["340", "742"]
```

jwhitlock:
I've split the first part, bringing `mobile-codes` into this project, into PR #1444. I think it will make this PR easier to comprehend.

I've spent the morning with the code, and the list of changes I want to make gets bigger and bigger. I think I'm going to capture those as follow-on issues, and work toward getting this updated data merged.

mobile_codes/json/countries.json
```diff
@@ -2,7 +2,7 @@
     ["Afghanistan", "AF", "AFG", "004", "412"],
     ["Albania", "AL", "ALB", "008", "276"],
     ["Algeria", "DZ", "DZA", "012", "603"],
-    ["American Samoa", "AS", "ASM", "016", "544"],
+    ["American Samoa", "AS", "ASM", "016", ["311", "544"]],
```
jwhitlock, line 5: How to read this:

* Country name: ``American Samoa``
* [ISO 3166-1 alpha-2 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2): ``AS``
* [ISO 3166-1 alpha-3 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3): ``ASM``
* [ISO 3166-1 numeric code](https://en.wikipedia.org/wiki/ISO_3166-1_numeric): ``016``
* [Mobile Country Codes](https://en.wikipedia.org/wiki/Mobile_country_code): 
  - ``311``: Generally United states, but used by American Samoa Telecommunications
  - ``544``: National operator, Bluesky in American Samoa

mobile_codes/json/countries.json
```diff
@@ -19,23 +19,22 @@
...
     ["Benin", "BJ", "BEN", "204", "616"],
-    ["Bermuda", "BM", "BMU", "060", ["350", "338", "310"]],
+    ["Bermuda", "BM", "BMU", "060", ["310", "338", "350"]],
     ["Bhutan", "BT", "BTN", "064", "402"],
     ["Bolivia, Plurinational State of", "BO", "BOL", "068", "736"],
     ["Bonaire, Sint Eustatius and Saba", "BQ", "BES", "535", "362"],
     ["Bosnia and Herzegovina", "BA", "BIH", "070", "218"],
     ["Botswana", "BW", "BWA", "072", "652"],
-    ["Bouvet Island", "BV", "BVT", "074", "242"],
+    ["Bouvet Island", "BV", "BVT", "074", null],
```
jwhitlock, line 31: [Bouvet Island](https://en.wikipedia.org/wiki/Bouvet_Island) is a dependency of Norway, and ``242`` is the MCC for Norway. I guess there is no cell reception there...

mobile_codes/json/mnc_operators.json
```diff
@@ -1,1344 +1,1933 @@
...
-    ["204", "05", "Elephant Talk Communications Premium \nRate Services", "Elephant Talk Communications Premium \nRate Services"],
-    ["204", "06", "Mundio Mobile (Netherlands) Ltd", "Mundio Mobile (Netherlands) Ltd"],
-    ["204", "07", "Teleena holding B.V.", "Teleena holding B.V."],
-    ["204", "08", "KPN Mobile The Netherlands B.V.", "KPN Mobile The Netherlands B.V."],
-    ["204", "09", "Lycamobile Netherlands Limited", "Lycamobile Netherlands Limited"],
-    ["204", "10", "KPN B.V.", "KPN B.V."],
+    ["001", "001", "TEST", "Test network"],
+    ["001", "01", "TEST", "Test network"],
+    ["202", "01", "Cosmote", "COSMOTE - Mobile Telecommunications S.A."],
+    ["202", "02", "Cosmote", "COSMOTE - Mobile Telecommunications S.A."],
```
jwhitlock, line 5: Previously, the brand and operator were generally the same string. These are now different, parsed from the Wikipedia source.

mobile_codes/json/mnc_operators.json
```diff
@@ -1,1344 +1,1933 @@
...
+    ["204", "06", "Vectone Mobile", "Mundio Mobile (Netherlands) Ltd"],
+    ["204", "07", "", "Teleena (MVNE)"],
+    ["204", "08", "KPN", "KPN Mobile The Netherlands B.V."],
+    ["204", "09", "Lycamobile", "Lycamobile Netherlands Limited"],
+    ["204", "10", "KPN", "KPN B.V."],
     ["204", "11", "", "VoipIT B.V."],
-    ["204", "12", "KPN B.V.", "KPN B.V."],
-    ["204", "13", "Unica Installatietechniek B.V.", "Unica Installatietechniek B.V."],
+    ["204", "12", "Telfort", "KPN Mobile The Netherlands B.V."],
+    ["204", "13", "", "Unica Installatietechniek B.V."],
```
jwhitlock, line 31: This means there are more blank brands, when the [Wikipedia source](https://en.wikipedia.org/wiki/Mobile_Network_Codes_in_ITU_region_2xx_(Europe)#Netherlands_(Kingdom_of_the_Netherlands)_–_NL) has an empty Brand column

mobile_codes/source_data/itu.json
```diff
@@ -1,1745 +1,2038 @@
...
-    ["901", "33", "Smart Communications , Inc", "Smart Communications , Inc"],
-    ["901", "34", "Tyntec GmbH", "Tyntec GmbH"],
-    ["901", "35", "Globecomm Network Services", "Globecomm Network Services"],
-    ["901", "36", "Azerfon LLC", "Azerfon LLC"],
-    ["901", "37", "TRANSATEL", "TRANSATEL"],
-    ["901", "38", "Multiregional TransitTelecom (MTT)", "Multiregional TransitTelecom (MTT)"],
-    ["901", "39", "MTX Connect", "MTX Connect"],
-    ["901", "40", "Deutsche Telekom AG", "Deutsche Telekom AG"],
-    ["901", "88", "UN Office for the Coordination of Humanitarian Affairs (OCHA)", "UN Office for the Coordination of Humanitarian Affairs (OCHA)"]
+    ["202", "01", "Cosmote AE", "Greece"],
```
jwhitlock, line 2: The format changed to:

* [Mobile Country Code](https://en.wikipedia.org/wiki/Mobile_country_code) (MCC) (_same_)
* Mobile Network Code (_same_)
* Brand / Operator (_same_)
* Country (was Operator / Brand again)

jwhitlock (commented):
Some notes on the changes, before I change the tests to pass.

jwhitlock:
I've gotten the tests working, and filed some follow-on issues for potential future work (issues #1445 and #1446). This is a big step forward in data coverage, and toward regular updates. Thanks again, @szjozsef!

mobile_codes/json/mnc_operators.json
szjozsef: Yes in many cases they was the same, because the main original source itu.json, (derived from the itu publication) does not have a distinct column for brand/operator, and it seems the original project just copied the same information in both fields. But used the wikipedia data as an updated over that initial import. The same is true also for the added additional data source mcc-mnc-table.json.

mobile_codes/json/mnc_operators.json
szjozsef: Yes. But if somehow (?) this is causes any issue the parser script can be adapted to use the operator value in case the brand is an empty string

mobile_codes/source_data/itu.json
szjozsef: Yes. The initial file contained the same value on the last 2 columns (the itu document does not have the brand value, just operator name), instead of having the same value twice, was more preferable to have also the country name, this way now the country information also can be imported from the itu sources as well.

Issue #1390, opened 2020-10-27
Update the node environment


jwhitlock:
Node is used in development to download resources such as mapbox.js as NPM packages, and to combine multiple resources into a single download. Node is not used in the build (combined assets are checked into the project) or in production.

The node environment is running on Node 8, and Node 14 is the current LTS release (check https://nodejs.org/en/about/releases/ to see if this is still accurate). The packages also appear out of date, and some (such as font awesome) may no longer be used in the build. The out-of-date mapbox.js package was responsible for MLS using a deprecated Mapbox server, causing concern for at least 3 ops staff (see [closed bug 1614130](https://bugzilla.mozilla.org/show_bug.cgi?id=1614130)).

This covers getting the node environment and packages up to current versions, updating any docs, and investigating if dependabot can be used to suggest updates in the future.

*Originally filed as [bug 1642372](https://bugzilla.mozilla.org/show_bug.cgi?id=1642372) on 2020-06-01*

jwhitlock:
Node is updated to 15.x, and Dependabot now scans node dependencies once a month. Some manual work may be needed to update ``npm-shrinkwrap`` and to rebuild CSS and JS when needed.

PR #1447, opened 2020-12-14
Update from node 8.x to 14.x

jwhitlock:
Update node from 8.x to 14.x (issue #1390), with no package updates. Enable Dependabot for the Dockerfile and for the npm packages.

Node is used to install javascript libraries like jquery and mapbox, and to combine and minify CSS and JS for the website. It is not run in production.


jwhitlock:
CI passes. Let's see if Dependabot can update the packages...

Issue #1440, opened 2020-12-11
Replace pytz.UTC with zoneinfo('UTC')


jwhitlock:
We're using ``pytz.UTC`` for timezone-aware dates. Python 3.9 has added [zoneinfo](https://docs.python.org/3.9/library/zoneinfo.html) as a standard module. If we use this, then ``pytz`` can be removed from first-level requirements.

Old code:

```
from pytz import UTC

def utcnow():
    """Return the current time in UTC with a UTC timezone set."""
    return datetime.utcnow().replace(microsecond=0, tzinfo=UTC)
```

New code:

```
from zoneinfo import ZoneInfo

def utcnow():
    """Return the current time in UTC with a UTC timezone set."""
    return datetime.utcnow().replace(microsecond=0, tzinfo=ZoneInfo("UTC"))
```

The timezone-aware ``datetime`` objects created with ``pytz.UTC`` are equivalent to the ones created with ``ZoneInfo("UTC")`` - see https://gist.github.com/jwhitlock/08ca74b037c2196f946262ef1a4b775d

dantesCode:
I want to make this issue :)

jwhitlock:
This was complete after #1462 and #1464, and should have been closed in December 2020. Thanks for your help @dantesCode!

PR #1462, opened 2020-12-17
Replace pytz.UTC with zoneinfo('UTC')

dantesCode:
[#1440] Replace pytz.UTC with zoneinfo('UTC')


dantesCode:
OK, I'll make another pull request.

PR #1468, opened 2020-12-18
parsing for British Indian Ocean Territory

szjozsef:
In addition to #1421 add an exceptional parsing for British Indian Ocean Territory.
This will replace the null mccs list in the countries.json for "British Indian Ocean Territory" to "995"


jwhitlock (approved):
Thanks @szjozsef!

PR #1470, opened 2020-12-23
split mnc ranges to individual entries

szjozsef:
For example ["314", "100 - 190", "", "Reserved for Public Safety"],
will get split into 91 individual entries, creating valid entries
like this:
["314", "100", "", "Reserved for Public Safety"],
....
["314", "190", "", "Reserved for Public Safety"],

The updated datafiles also contains some changes made to the wikipedia pages for US Virgin Islands (I hope that changes will not get reverted ...)


szjozsef:
I see that this tests are failing:

>  File "/app/ichnaea/tests/test_geocode.py", line 78, in test_multiple
    assert set(GEOCODER.regions_for_mcc(311)) == {"AS", "GU", "US"}
AssertionError: assert {'AS', 'GU', 'US', 'VI'} == {'AS', 'GU', 'US'}
  Extra items in the left set:
  'VI'

Because the failed test is in geocoder part I don't want just to change the test to pass. But maybe it is OK to adapt the test case.
I would like to somebody double check the changes to the https://en.wikipedia.org/wiki/Mobile_Network_Codes_in_ITU_region_3xx_(North_America) pages regarding the United States Virgin Islands networks. That information entered there was a result to what I can found about Mobile networks in the USVI, inclusiv the interactive map from https://opencellid.org . I found a lot of reported cell's with MCC 311 MNC 470 and MCC 310 MNC 410 (from time period ~ 2015 - 2018) So the 310 and 311 seems pretty valid for USVI. I have no clue if the Liberty re-branding of the AT&T network there also imply some code changes or not.

szjozsef:
It seems that with the changed test case for GEOCODER.regions_for_mcc(311) all the tests are passed.
This test seems a little fragile to me, because some data changes can trigger test failures.

jwhitlock:
I found this [press release](https://about.att.com/story/2020/att-liberty-latin-america.html) that says AT&T sold its Puerto Rico and U.S. Virgin Island operations to [Liberty Latin America](https://lla.com), so it is changing carriers instead of re-branding.

jwhitlock (approved):
Thanks, @szjozsef, the changes look good. I think it is still OK for tests to fail when the data changes, but I'd support testing a mobile provider other than "311" if it continues to rapidly change.

PR #1479, opened 2021-01-12
Update datamaps script and docs

### LINKED ISSUE #1404: Document the data flow for Datamap data
Fill out the details of populating [Datamap data](https://ichnaea.readthedocs.io/en/latest/dataflows.html#stat-data).

jwhitlock:
Document and rewrite ``ichnaea/scripts/datamap.py``, the script that generates map tiles of observation points and uploads them to an S3 bucket. This fixes #1404 (document the data flow for datamap tiles) and addresses issue #840 (generate datamap tiles again).

## Command line changes
    
* The script is now executable.
* Check that the AWS credentials allow reading, writing, and deleting objects in the S3 bucket at the start of the script, rather than failing at the start of upload.
* Allow uploading of previously created tiles.
* Check that a pre-existing ``--output`` folder is used when only creating or uploading tiles. This avoids creating tiles in a temporary folder that is then deleted, or deleting all the S3 tiles by syncing an empty temporary folder.
* Use [os.sched_getaffinity(0)](https://docs.python.org/3/library/os.html#os.sched_getaffinity) for default concurrency when available, for more accurate CPU count than [os.cpu_count()](https://docs.python.org/3/library/os.html#os.cpu_count) in Docker images and some Linux systems.
* Capture tool output and replace with logging-based progress messages.

## Functionality changes
    
* Remove timer and counter metrics.
* Add structured logging, with execution details (such as timers) in the structured data.
* Add execution details to ``tiles/data.json`` in the S3 bucket.
* Do not gzip CSV files. These now require 3x more disk space, but the size is still dwarfed by the generated tile images.
    
## Code refactoring
    
* Rename functions, such as changing ``encode_file()`` to ``csv_to_quadtrees()``.
* Rearrange functions to place ``generate()`` near top of file.
* Switch from celery's [billiard](https://github.com/celery/billiard) to standard library's [multiprocessing](https://docs.python.org/3/library/multiprocessing.html).
* Switch from [os.system()](https://docs.python.org/3/library/os.html#os.system) to [subprocess](https://docs.python.org/3/library/subprocess.html). There are minor speed gains from avoiding shell initialization, but the real goal is to prepare for [asyncio.create_subprocess_exec()](https://docs.python.org/3/library/asyncio-subprocess.html).
* Split up functions for self-documentation, and to break up shell pipes into separate steps to allow progress reporting and future chaining with ``asyncio``.
* Sync entire tiles folder to S3 rather than break up into syncing subfolders, and fully determine the sync actions before starting syncing. This may slightly increase sync time, but will hopefully work better in a future ``asyncio`` workflow.
    
## Tests
    
* Refactor tests to use ``pytest`` fixtures for test setup.
* Expand tests for the new ways to invoke the script.

## Documentation
    
* Write the data flow section for datamap updates.
* Rewrite the datamap section of the metrics doc. The ``datamap.dberror`` metric belongs with the task app and internal database updates. The script no longer emits timing or counter metrics, but instead outputs a detailed ``canonical-log-line``.
* Expand configuration documents around map and tile configuration.
* Add documentation for generating tiles for the development environment, and for using an S3 bucket for tiles.

## Webapp changes

* Change the default ``ASSET_URL`` from an empty string (for a tiles URL of ``/tiles``) to ``/static/datamap/`` (for a tiles URL of ``/static/datamap/tiles``). This allows locally generated tiles to be viewed in the dev environment. Production will continue to set ``ASSET_URL`` to the S3 bucket's public endpoint.

## Other

* Remove ``MANIFEST.in``, which should have been removed with ``setup.py`` in PR #1427.
* Move the local tile folder for development from ``ichnaea/content/static/tiles/tiles`` ``ichnaea/content/static/datamap/tiles``.
* Update ``.dockerignore`` to sync with changes after 1.5, and to ignore the development tiles folder when building images.
* Update ``.gitignore`` for the new tile folder ``ichnaea/content/static/datamap``.
* Add ``make local-map`` to generate local tiles.


jwhitlock:
I've revised the code and docs a few times, and ran the code successfully in the local dev environment. The next step is manual tests in the staging environment.

PR #1481, opened 2021-01-14
Datamap script: Limit process usage, handle Ctrl-C, more logging tweaks

jwhitlock:
For issue #840, switching to the low-level S3 client in PR #1480 did not help the out-of-memory messages. Next, I'm trying [maxtasksperchild](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool), hoping that destroying and recreating the task thread will reset memory usage.

Hitting Ctrl-C to halt the script was causing a ``KeyboardInterrupt`` exception to be logged in Sentry. This adds a handler for this event to skip logging to Sentry, and adjusts the final log message if it halted due to an interrupt.

This adds debug log messages for the start of tile creation and upload. Otherwise, there is a long period with no feedback that the script is running.

This updates a few debug log messages to fix pluralization and to more consistently use commas as thousands separators. 

To further reduce memory usage:
* Track the number of unchanged files, rather than a list of filenames of unchanged files
* Remove the S3 client session after fetching the metadata for existing files

### DIFF
--- a/ichnaea/scripts/datamap.py
+++ b/ichnaea/scripts/datamap.py
@@ -233,6 +233,9 @@ def watch_jobs(
                 job_resp = job.get()
                 if on_success:
                     on_success(job_resp)
+            except KeyboardInterrupt:
+                # Skip Raven for Ctrl-C, reraise to halt execution
+                raise
             except Exception as e:
                 if raven_client:
                     raven_client.captureException()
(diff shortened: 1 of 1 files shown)


jwhitlock:
This ran to completion on stage on an instance with 1 CPU and 2 GB memory in 5 hours. Rendering took 3.5 hours, and upload took over an hour.  The changes appeared to drop memory usage to around 300 MB, but it is difficult to tell with shared memory and threads.

PR #1483, opened 2021-01-20
Datamap script updates for production

jwhitlock:
**Work in Progress** - These changes are being tested on a production server.

For issue #840, make further changes to the datamap script and documentation:

* When an export CSV reached 10,000,000 rows, start writing to multiple CSV files. These are processed into separate quadtree files, which are then merged (duplicates allowed) into the per-table quadtree files, before being merged (eliminating duplicates) into a final whole-world quadtree. This should avoid out-of-memory errors when converting a very large tables from CSV to a quadtree.
* Adjust database mocking method, so that we can test for the ``generate()`` function and the serialized CSV paths.
* Update documentation for new structured logging data.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -905,8 +905,11 @@ The final ``canonical-log-line`` log entry has this data:
 * ``bucket_name``: The name of the S3 bucket
 * ``concurrency``: The number of concurrent threads used
 * ``create``: True if ``--create`` was set to generate tiles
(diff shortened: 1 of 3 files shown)


jwhitlock:
This worked in production, but I did notice that ``tiles_failed`` was recorded but not documented, and ``tiles_unchanged`` was no longer recorded, so I added those as well.

PR #1585, opened 2021-05-17
Fix minor docs typo: WSG 84 -> WGS 84

personalcomputer:
Just fixing a minor typo. The correct term for the coordinate system used by GPS is **WGS** 84, standing for World Geodetic System, not "WSG" 84. reference: https://en.wikipedia.org/wiki/World_Geodetic_System


jwhitlock (approved):
Thank you @personalcomputer!

Issue #1636, opened 2021-07-07
Make development environment compatible with Apple Silicon M1


jwhitlock:
Apple computers released in 2020 and later come with the M1 chip, which is detected as `linux/arm64/v8` by [Docker Desktop for Apple Silicon](https://docs.docker.com/docker-for-mac/apple-silicon/). Previous computers were detected as `linux/amd64`. The application image is unable to build on a M1 laptop, because MySQL does not distribute `arm64` packages, only `amd64` and `i386`. There may be further issues after this is solved.

Some possibilities:

* Run under emulation, with ``--platform linux/amd64`` and other options
* Switch to [MariaDB](https://mariadb.com/database-topics/mariadb-vs-mysql/) for client libraries, which should be compatible
* Switch from MySQL to MariaDB for the database server as well.

We intend to continue running MySQL 5.7 in production, and want to avoid issues due to further mismatches between development and deployments.

jwhitlock:
I've read and experimented a bit, and got to working code on my M1-based laptop.

The Oracle MySQL team [supports MySQL server 8.0](https://hub.docker.com/r/mysql/mysql-server/) with x86 and arm64 releases. I don't expect them to ever support arm64 for 5.7.

I haven't researched why, but the official Debian package repositories stop at MySQL 5.5, so we use a repository hosted by mysql.com. The MySQL Debian repo does not have arm64 packages for [MySQL 5.7](https://repo.mysql.com/apt/debian/dists/buster/mysql-5.7/) or [MySQL 8.0](https://repo.mysql.com/apt/debian/dists/buster/mysql-8.0/). There may be something that changed recently, since there is a sid (unstable) build of [MySQL server 8.0.23-3+b1](https://packages.debian.org/sid/mysql-server-8.0) for arm64 and other architectures.

I don't think waiting for upstream support will be an effective option. It may be possible to run a MySQL 8.0 server image under arm64 locally, but that should be part of a production server update from 5.7 to 8.0, and doesn't address the client libraries.

I tried running the MySQL 5.7 docker image with emulation, but it failed to load. It is possible that I could get it working with more effort, or by turning on some experimental flags. I'm not excited by the [docker notes](https://docs.docker.com/docker-for-mac/apple-silicon/), which mention crashes, missing features, and “best effort” for container emulation.

[...]

PR #1637, opened 2021-07-08
Switch to MariaDB client libraries

jwhitlock:
MySQL [does not provide arm64 binaries](https://repo.mysql.com/apt/debian/dists/buster/mysql-5.7/), but MariaDB client libraries, which [provide a compatible interface](https://mariadb.com/database-topics/mariadb-vs-mysql/), provides arm64 and amd64 binaries and are suggest by `apt-get`. This will allow the same library to be used in development on Intel and Apple Silicon, as well as in production deployments.

Addresses issue #1636.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,15 +23,6 @@ path-exclude=/usr/share/man/*\n\
 path-exclude=/usr/share/locale/*\n\
 " > /etc/dpkg/dpkg.cfg.d/apt-no-docs
 
-# Install GNUPG.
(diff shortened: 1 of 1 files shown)


jwhitlock:
Tests pass in CI, and on my Intel-based laptop. Waiting to merge until after next production push.

jwhitlock:
On second thought, I'd rather have this code running on stage. It should either run without any noticeable change, or expose issues that we'll only see in a deployed environment.

Issue #1323, opened 2020-09-22
"make setup" fails if MySQL is not ready


jwhitlock:
When [setting up a local development environment](https://ichnaea.readthedocs.io/en/latest/local_dev.html), the sequence of commands is:

```shell
make runservices  # Start Redis and MySQL
make setup # Initialize the location database
```

``make setup`` will fail if MySQL is still initializing. Ideally, either ``make runservices`` would wait until the services are initialized, or ``make setup`` will wait until the MySQL port is available to run commands.

PR #1661, opened 2021-08-13
Add option to switch database to MariaDB 10.5

jwhitlock:
MySQL 5.7 does not provide the arm64 images needed to run on Apple Silicon M1s on the the 2020 MacBooks (issue #1636). MariaDB 10.5 does provide them, and should be compatible with MySQL 5.7.

In the development environment:

* Use ``ICHNAEA_DOCKER_DB_ENGINE`` to switch between ``mysql_5_7`` (the default) and ``mariadb_10_5``
* Rename the database container from ``mysql`` to ``db``
* ``make build`` now builds the ``db`` and ``redis`` containers
* ``make mysql`` is now ``make dbshell``, to get a interactive SQL shell
* Update documentation to mention ``ICHNAEA_DOCKER_DB_ENGINE``, and how to wait for the database to initialized (addresses issue #1323)


jwhitlock:
I'm not thrilled with this solution, but 1) it is dev only, and 2) I'm basically the only dev, and 3) I need to get these monthly updates tested and pushed.

I hope to add MySQL 8.0 to the mix, unless I suddenly get a ton of MLS time and can add PostgreSQL.

PR #1762, opened 2022-01-05
Create cross-platform docker images

jwhitlock:
For issue #1636, create Docker images with `linux/amd64` and `linux/arm64` variants.  This uses techniques described on the tutorial [Building Docker images for multiple operating system architectures](https://circleci.com/blog/building-docker-images-for-multiple-os-architectures/).

This add ``make cross-build`` to run ``docker buildx`` to build the multi-platform images, but doesn't ensure that ``buildx`` is installed and configured.

The CircleCI configuration is changed to use a machine image, using similar but slightly different steps.  The bundled ``docker`` is used, and version details are logged and added to an artifact.  A ``when`` step is used to determine if we're building a tag or the ``main`` branch. If this is the case, then ``buildx`` is installed and configured, and used to build the image tagged with the SHA1 / tag name, as well as the ``latest`` tag if it is the main branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,8 @@ commands:
             docker-compose --version | tee -a << parameters.info_path >>
             echo "*** docker compose --version ***" | tee -a << parameters.info_path >>
             docker compose --version | tee -a << parameters.info_path >>
+            echo "*** docker buildx inspect --bootstrap ***" | tee -a << parameters.info_path >>
+            docker buildx inspect --bootstrap | tee -a << parameters.info_path >>
       - store_artifacts:
           path: << parameters.info_path >>
           destination: << parameters.artifact_path >>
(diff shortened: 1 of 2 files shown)


jwhitlock:
I've tested `make cross-build` locally, but it doesn't run in CircleCI yet. I'll see if I can adjust the condition to test against this branch...

jwhitlock:
This worked, but it too 30 minutes to build the two variants, versus about 5 minutes for just the amd64 version. I don't think this will be acceptable, since the arm64 version is mostly for developers.

I'm curious if an arm64 server would be any faster:

https://circleci.com/docs/2.0/arm-resources/

It may make sense to go back to a previous version of this command, where all the build / lint / test steps were in the same job, and then run the same job on an arm64 machine, and then build the manifest in a follow-on step.

This may require using `docker manifest`, described here as the "hard way":

https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

It may also require using workflows to get the data from the machine-specific builds to the manifest writer:

https://circleci.com/docs/2.0/workflows/#using-workspaces-to-share-data-among-jobs

Of course, now we're approaching the effort to migrate to Taskcluster...

jwhitlock:
I'm going to close this PR. I don't think the end result is worth the effort at this time. As far as I know, 100% of the people that want to run `arm64` are me, and `make build` works fine for that.

PR #1779, opened 2022-02-01
Downgrade to Shapely 1.7.1

jwhitlock:
Shapely 1.8.0 has packaging issues which prevent building with speedups (see https://github.com/shapely/shapely/issues/1263). Wait for Shapely 1.8.1 to update.


jwhitlock:
Downgrading to Shapely 1.7.1 works with Python 3.9 but not 3.10 😢.

PR #1780, opened 2022-02-01
Dependency updates, February 2022, Part 1 - Language updates

jwhitlock:
Disable speedups check for Shapely 1.8.0. This is used for region from lat/long estimate, which is not currently used by any clients.

Then, update the language versions:

* Bump python from 3.9.9-slim to 3.10.2-slim (from PR #1765)
* Bump node from 16.13.1-slim to 16.13.2-slim in /docker/node (from PR #1766)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9.9-slim
+FROM python:3.10.2-slim
 
 # Set up user and group.
 ARG groupid=10001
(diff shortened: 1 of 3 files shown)


jwhitlock:
Looking at the code, the `GEOCODER` may be used in some additional paths, when attempting to determine the country for a station. This means that there my be a performance drop without speedups. Putting Python 3.10 on the shelf until Speedups 1.8.1 is released.

PR #1934, opened 2023-01-19
Update codeowners

jwhitlock:
Update the code owners so that @bsieber-mozilla is the default reviewer.


bsieber-mozilla (approved):
thanks @jwhitlock

PR #1949, opened 2023-04-22
Update scipy, Python 3.11, Redis

jwhitlock:
Manually update some dependencies and the Python version, where Dependabot is having trouble.

* pip-tools 6.13.0 - This is the latest version, and works with pip 23.1, which changed some interfaces
* scipy 1.10.1 - Once pip-tools is updated, it can bring in the new hashes for the pre-compiled wheels, avoiding issues trying to build it from source
* gevent 22.10.2 - The latest gevent supports Python 3.11 and makes the leap to greenlet 2.0
* Python 3.11.3 - The headline feature is a 10% - 60% speedup over Python 3.10. It also includes a built-in TOML standard library, so we no longer need `tomli`.
    - Because Python is faster, the tests now start running before MySQL is ready. Added code to confirm that MySQL is ready, and sleep for 5 seconds if not, giving up after 30 seconds. With Python 3.11, a 5 second nap is sufficient. 
* redis 4.5.4 - After all those changes, the redis update is smooth. It doesn't need `async-timeout` with Python 3.11, so `geoip2` can keep its preferred version.


bsieber-mozilla (approved):
Thank you for this PR!

Issue #2043, opened 2023-10-18
Data loss when uploading measurements but API responds OK


zamojski:
I'm developer of [Tower Collector](https://github.com/zamojski/TowerCollector/) app which contributes to MLS. One of the users from Japan had recently reported an issue related to data loss. Below quote from the email.

> There has been a problem with the Mozilla Location Service since about July 2022. The problem is that the Mozilla Location Service sometimes ignores the data collected by Tower Collector when it is uploaded.
> 
> When this problem occurs, the geosubmit API of the Mozilla Location Service is still responding with HTTP 200 OK and "{}" data, so the
> upload appears to have completed successfully in Tower Collector. We are confident this is not a Tower Collector issue.
> 
> In some cases, the recent Mozilla Location Service is behaving in a way that even if the Upload is successful, the data parsing times out
> and discards the data uploaded by the user. (I haven't checked properly, but there is about a 70% chance that the data is discarded.)
> 
>...
>
> If the data does not appear in the Mozilla Location Service after a long wait (about one day) after uploading, we assume that the data has been discarded.

@jwhitlock or someone well informed can take a look on that? It seems to be significant issue.

jwhitlock:
Thanks for the report @zamojski, and I'm sorry the service is not working as expected. I'm no longer on the team supporting this codebase or MLS, but I can give some context.

Looking at the submission code:

https://github.com/mozilla/ichnaea/blob/47d34a57a3b1b39728a2fe064e81690d2dcdbdb0/ichnaea/api/submit/views.py#L41-L75

There's a few things that can go wrong. If the data is omitted, a 400 is returned. If our caching server is down, a 502 is returned. If the API key is wrong, a 200 OK is still returned. They should double-check the API key, such as seeing if it is valid on /v1/geolocate.

The fact that _some_ are getting through could mean a processing issue. The submission is queued and processed on the backend, not during the request. The processing time is uncertain, but I'd expect within 24 hours. If the system gets overloaded, it does drop some observations. Submissions are much less likely to be dropped than other observations.

isogame3:
Hello. I am the user who consulted @zamojski about the data loss issue at the beginning of this thread.

Recently, I conducted an experiment.
I collected 24 eNB-LCID measurements from cell towers in areas where few people visit.
[...]

PR #2062, opened 2024-01-12
Removing 1 second delay for data tasks that are ready to be processed

alexcottner:
Relates to issue https://github.com/mozilla/ichnaea/issues/2043 - the data queues can get backed up causing data to be lost. 

This removes an unnecessary 1 second delay between processing batches of records. On the surface this doesn't seem like much, but in local testing it makes a huge difference and will hopefully alleviate the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ ichnaea/content/static/datamap/tiles/
 
 my.env
 .docker-build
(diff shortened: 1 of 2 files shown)


jwhitlock (approved):
Thanks @alexcottner, looks work trying to me!

PR #2067, opened 2024-07-30
Fix typo in README

fmarier:


fmarier:
cc @alexcottner

alexcottner (approved):
Thanks for catching that

